PR #14, opened 2016-04-07
Added empty otherProperties view and GET method.

jameshforster:


jameshforster:
Didn't include model.

PR #16, opened 2016-04-07
CGT-674 added annual-exempt-ammount model, tests, action and html.

DavidWoodward37:
Don't deal with this until I have had a chance to maybe fix a master merge conflict.

### DIFF
--- a/app/models/AnnualExemptAmountModel.scala
+++ b/app/models/AnnualExemptAmountModel.scala
@@ -18,7 +18,7 @@ package models
 
 import play.api.libs.json._
 
-case class AnnualExemptAmountModel (customerType: String)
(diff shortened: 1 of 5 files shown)


app/models/AnnualExemptAmountModel.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+
+import play.api.libs.json._
+
+case class AnnualExemptAmountModel (customerType: String)
```
jameshforster: Should be annualExemptAmount and should be type BigDecimal

PR #19, opened 2016-04-07
CGT-687 added model, view, test and updated controller for disposal value

DavidWoodward37:


MJCallahanPage:
Dave, looks like master might need to be merged into this branch.

PR #31, opened 2016-04-08
Added href to button on introduction page and called reverse routing to next page.

jameshforster:


app/views/cgts/introduction.scala.html
```diff
@@ -8,7 +8,7 @@ <h1 class="heading-large">@Messages("calc.introduction.header")</h1>
     <p>@Messages("calc.introduction.textOne")</p>
     <p>@Messages("calc.introduction.textTwo")</p>
 
-    <button id="start" class="button button-start">@Messages("calc.introduction.start")</button>
+    <a id="start" href=@controllers.routes.CalculationController.customerType class="button button-start">@Messages("calc.introduction.start")</a>
```
DavidWoodward37: Checked this against nisp and it seems to be right but do you know the reason for not using the buttons or is it just because it 'wont work?'

PR #32, opened 2016-04-08
CGT-682 added tests and html content with messages for disposal date.

DavidWoodward37:
Also updated the helper for a form date.


app/views/cgts/disposalDate.scala.html
```diff
@@ -1,6 +1,17 @@
...
 
-@views.html.main_template(title = "Disposal Date", bodyClasses = None) {
-    @simpleInlineDate("dod")
+@views.html.main_template(title = Messages("calc.disposalDate.title"), bodyClasses = None) {
+    <div class="phase-banner-alpha">
+        <p>
+            <strong class="phase-tag">ALPHA</strong>
+            <span>This is a new service – your <a href="https://designpatterns.hackpad.com/GOV.UK-elements-feedback-sKrDyQxcfA2">feedback</a> will help us to improve it.</span>
+        </p>
+    </div>
```
MJCallahanPage: Dave, there is a betaBanner section in the govuk_wrapper.scala.html that we probably need to cofigure for this so that it is consistent across all pages. 

I suggest we create this as a separate task.

Otherwise, we would have to remove this from each individual page at a later date rather than just in one place.

app/views/cgts/disposalDate.scala.html
```diff
@@ -1,6 +1,17 @@
...
+    <a class="link-back" href="javascript: history.go(-1)">@Messages("calc.base.back")</a>
```
MJCallahanPage: This link, using javascript will cause issues for people that don't have javascript enabled - it can also perform some wierd things sometime.

Can you use the reverse routing feature of scala. E.g.:

`href=routes.CalculationController.improvements`

So that it is the get of the previous page.

app/views/cgts/disposalDate.scala.html
```diff
@@ -1,6 +1,17 @@
...
+    <h1 class="heading-xlarge"><span class="heading-secondary">@Messages("calc.base.calculate.tax.heading")</span>25% @Messages("calc.base.percentage.comp")</h1>
+
```
MJCallahanPage: As the user story doesn't mention the percentage complete I think this should be removed. This will be re-introduced if it is confirmed as needed as part of a separate story.

app/views/cgts/disposalDate.scala.html
```diff
@@ -1,6 +1,17 @@
...
+    @simpleInlineDate("dod", Messages("calc.disposalDate.title"))
+
+    <button id="start" class="button button-start">@Messages("calc.base.button.continue")</button>
```
MJCallahanPage: This isn't a start button. Can you change so that it is more relevant to continue

app/views/cgts/disposalDate.scala.html
```diff
@@ -1,6 +1,17 @@
 @()(implicit request: Request[_])
 @import views.html.helpers._
 
-@views.html.main_template(title = "Disposal Date", bodyClasses = None) {
-    @simpleInlineDate("dod")
+@views.html.main_template(title = Messages("calc.disposalDate.title"), bodyClasses = None) {
```
MJCallahanPage: You could just use `@main_template` here and it should still work.

app/views/helpers/simpleInlineDate.scala.html
```diff
@@ -14,28 +14,28 @@
...
 @dayElemId = @{fieldName + ".day"}
 @monthElemId = @{fieldName + ".month"}
 @yearElemId = @{fieldName + ".year"}
 
-<fieldset class="form-group" id="dod">
-    <legend class="visuallyhidden">Date you signed the contract that made someone else the owner</legend>
-    <span class="form-hint">For example, 4 9 2016</span>
+<fieldset class="form-group" id="@fieldName">
+    <legend>@legendMessage</legend>
+    <span class="form-hint">@Messages("calc.date.hint")</span>
```
MJCallahanPage: Would the hint for the date always be the same? Or would this be better as a parameter to the helper?

app/views/cgts/disposalDate.scala.html
```diff
@@ -1,6 +1,17 @@
...
+    <div class="phase-banner-alpha">
+        <p>
+            <strong class="phase-tag">ALPHA</strong>
+            <span>This is a new service – your <a href="https://designpatterns.hackpad.com/GOV.UK-elements-feedback-sKrDyQxcfA2">feedback</a> will help us to improve it.</span>
+        </p>
+    </div>
+    <a class="link-back" href="javascript: history.go(-1)">@Messages("calc.base.back")</a>
+    <h1 class="heading-xlarge"><span class="heading-secondary">@Messages("calc.base.calculate.tax.heading")</span>25% @Messages("calc.base.percentage.comp")</h1>
+
+    @simpleInlineDate("dod", Messages("calc.disposalDate.title"))
```
MJCallahanPage: Is `dod` clear enough? I've seen `dod` used for "Date of Death" in most other HMRC implementations. So this might cause future confusion.

Could we just have `disposalDate`? So that the ids of the helper would then be rendered as `disposalDate.day` etc.

conf/messages
```diff
@@ -1,10 +1,15 @@
...
+calc.base.button.continue = Continue
 
 ## Common ##
-date.fields.day = Day
-date.fields.month = Month
-date.fields.year = Year
+calc.date.fields.day = Day
+calc.date.fields.month = Month
+calc.date.fields.year = Year
+calc.date.hint = For example, 4 9 2016
```
MJCallahanPage: How is "common" different to "base". Should we just have one section for common/base page items? If there is a need for common then should the names follow the existing pattern. E.g.:

`calc.**common**.date.day` et al.

test/controllers/CalculationControllerSpec.scala
```diff
@@ -122,9 +123,30 @@ class CalculationControllerSpec extends UnitSpec with WithFakeApplication {
...
+
+    "display the title from the messages file" in new fakeRequestTo("disposal-date") {
+      val result = CalculationController.disposalDate(fakeRequest)
+      contentAsString(result) should include (Messages("calc.disposalDate.title"))
+    }
+
+    "contain a simpleInlineDate with title When did you sign the contract that made someone else the owner?" in new fakeRequestTo("disposal-date") {
+      val result = CalculationController.disposalDate(fakeRequest)
+      contentAsString(result) should include ("<legend>" + Messages("calc.disposalDate.title") + "</legend>")
+    }
```
MJCallahanPage: You could use jsoup here to find particular ids to interrogate and enable the test to be very targeted and specific rather than scrapping the entire HTML body?

test/controllers/CalculationControllerSpec.scala
```diff
@@ -122,9 +123,30 @@ class CalculationControllerSpec extends UnitSpec with WithFakeApplication {
...
+
+    "contain a button with Continue on it" in new fakeRequestTo("disposal-date") {
+      val result = CalculationController.disposalDate(fakeRequest)
+      contentAsString(result) should include (Messages("calc.base.button.continue") + "</button>")
```
MJCallahanPage: Same here, could use jsoup maybe? :)

app/views/helpers/simpleInlineDate.scala.html
```diff
@@ -14,28 +14,28 @@
 * limitations under the License.
 *@
 
-@(fieldName: String)(implicit lang: play.api.i18n.Lang)
+@(fieldName: String, legendMessage: String)(implicit lang: play.api.i18n.Lang)
```
MJCallahanPage: Shouldn't we be passing in the form here so that we can reference the fields when creating the inputs? That way it will be easier to bind the form to the model?

app/views/helpers/simpleInlineDate.scala.html
```diff
@@ -14,28 +14,28 @@
...
+    <legend>@legendMessage</legend>
+    <span class="form-hint">@Messages("calc.date.hint")</span>
 
     <div class="form-date">
-        <div id="@dayElemId" class="form-group form-group-day">
-            <label for="dodDay">Day</label>
-            <input class="form-control" id="dodDay" name="dodDay" type="number" pattern="[0-9]*" min="0" max="31">
+        <div class="form-group form-group-day">
+            <label for="@dayElemId">@Messages("calc.date.fields.day")</label>
+            <input class="form-control" id="@dayElemId" name="@dayElemId" type="number" pattern="[0-9]*" min="0" max="31">
```
MJCallahanPage: You could potentially use the exiting gov.uk.helper for rendering input fields to help construct these, it would accept the field from the form and bind it to the input if you were to pass the form as a parameter to the helper.

app/views/helpers/simpleInlineDate.scala.html
MJCallahanPage: Form is a later task, so assumption is this will change then

PR #39, opened 2016-04-11
Added simple input box helper for monetary values.

jameshforster:


app/views/helpers/simpleInputMoney.scala.html
```diff
@@ -0,0 +1,10 @@
+@(fieldName: String, questionText: String, helpText: Option[String])(implicit lang: play.api.i18n.Lang)
+
+<fieldset class="form-group">
+    <label>@questionText</label>
```
DavidWoodward37: I don't think we changed this in the hotfix so do you want to fix it now? Needs to be a legend.

app/views/helpers/simpleInputMoney.scala.html
```diff
@@ -0,0 +1,10 @@
+@(fieldName: String, questionText: String, helpText: Option[String])(implicit lang: play.api.i18n.Lang)
```
DavidWoodward37: Default value here too?

PR #40, opened 2016-04-11
CGT-703: Disabled trustee basic html content markup and tests

MJCallahanPage:
**note**
The CalcuationControllerSpec has also been significantly changed. The fakeRequestTo class has been udpated to take the action of a controller as a parameter and then get the result and jsoupDoc from that result.

For the tests, we instantiate an object which extends this class passing the URL and the Action. We can then interrogate the values on the Object in our Scala Tests. There are methods that can be run on the Object to carry out generic tests. E.e. status200() checks for a 200 response.

This reduces the amount of duplicated code we have to write.


MJCallahanPage:
Closing as removing the changes made for the scala tests - team will look into this as separate task.

PR #43, opened 2016-04-12
CGT-703: Disabled trustee html content

MJCallahanPage:


app/views/cgts/disabledTrustee.scala.html
```diff
@@ -0,0 +1,14 @@
+@()(implicit request: Request[_])
+@import views.html.helpers._
+
+@main_template(title = Messages("calc.disabledTrustee.title")) {
```
DavidWoodward37: Can you remove the 'title =' part? It's not required as every other parameter has a default option if it is not supplied.

conf/messages
```diff
@@ -49,6 +49,8 @@ calc.customerType.personalRep = I was the executor of an estate
 
 ## Disabled Trustee ##
 calc.disabledTrustee.title = Are you a trustee for someone who’s vulnerable?
+calc.disabledTrustee.question = Are you a trustee for someone who’s vulnerable?
```
DavidWoodward37: Is there a point in having the same statement twice? Or should we re-name all the 'title' messages to questions as that's what they are?

PR #48, opened 2016-04-13
Test message refactor

DavidWoodward37:


DavidWoodward37:
I think there is an issue with the introduction page. Tests have been left as they were but they don't test all the message that are used on the page, and neither do they actual try to see if the UUID of the session is actually there.  Need to update..

PR #70, opened 2016-04-18
Added post method for Customer Type.

jameshforster:


test/controllers/CalculationControllerSpec.scala
```diff
@@ -339,7 +402,7 @@ class CalculationControllerSpec extends UnitSpec with WithFakeApplication with M
       object OtherPropertiesTestDataItem extends fakeRequestToPost(
         "allowance",
         TestCalculationController.submitOtherProperties,
-        ("otherProperties", "No")
+        ("otherProperties", "Yes")
```
MJCallahanPage: Why has other properties changed in this?

test/controllers/CalculationControllerSpec.scala
```diff
@@ -155,6 +155,69 @@ class CalculationControllerSpec extends UnitSpec with WithFakeApplication with M
     }
   }
 
+  "In CalculationController calling the .submitCustomerType action" when {
+    def keystoreCacheCondition[T](data: CustomerTypeModel): Unit = {
```
MJCallahanPage: Can this be pulled and made generic so that it is at the top of the class, rather than being repeated in each test section.

e.g. `def keystoreCacheCondition[T](data: Option[T]) ....... `

PR #83, opened 2016-04-19
CGT-724 added the post and model changes to the improvements page.

DavidWoodward37:


app/controllers/CalculationController.scala
```diff
@@ -158,6 +158,17 @@ trait CalculationController extends FrontendController {
     }
   }
 
+  val submitImprovements = Action { implicit request =>
+    improvementsForm.bindFromRequest.fold(
+      errors => BadRequest(calculation.improvements(errors)),
+      success => {
+        calcConnector.saveFormData("isClaimingImprovements", success.isClaimingImprovements)
```
jameshforster: Should save the entire model in a single entry rather than individual components.

test/controllers/CalculationControllerSpec.scala
```diff
@@ -826,6 +826,35 @@ class CalculationControllerSpec extends UnitSpec with WithFakeApplication with M
...
+        status(ImprovementsTestDataItem.result) shouldBe 303
+      }
+    }
+
+    "submitting an invalid form with 'Yes' and a value of 'fhu39awd8'" should {
+      object ImprovementsTestDataItem extends fakeRequestToPost("improvments", TestCalculationController.submitImprovements, ("isClaimingImprovements", "Yes"), ("improvementsAmt", "fhu39awd8"))
+      //This model actually has no bearing on the tes but the cachemap it produces is required.
+      val improvementsTestModel = new ImprovementsModel("Yes", Some(9878))
+
+      "return a 303" in {
```
jameshforster: Text does not match test.

PR #87, opened 2016-04-19
Added post method for Entrepreneurs Relief.

jameshforster:


test/controllers/CalculationControllerSpec.scala
```diff
@@ -1300,6 +1300,41 @@ class CalculationControllerSpec extends UnitSpec with WithFakeApplication with M
...
+        "entrepreneurs-relief",
+        TestCalculationController.submitEntrepreneursRelief,
+        ("entrepreneursRelief", "yes")
+      )
+      val testModel = new EntrepreneursReliefModel("Yes")
+
+      "return a 303" in {
+        keystoreCacheCondition[EntrepreneursReliefModel](testModel)
+        status(EntrepreneursReliefTestDataItem.result) shouldBe 303
+      }
```
MJCallahanPage: You could add a test here that proves that the page redirects to the correct next page.

e.g. An example from Disabled Trustee

```
s"redirect to ${routes.CalculationController.disposalValue()}" in {
        keystoreCacheCondition[DisposalDateModel](testModel)
        redirectLocation(DisposalDateTestDataItem.result) shouldBe Some(s"${routes.CalculationController.disposalValue()}")
      }
```

PR #92, opened 2016-04-19
CGT-710: Disposal date post route and validation

MJCallahanPage:


app/common/Validation.scala
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package common
+
+import java.text.SimpleDateFormat
+
+object Validation {
```
DavidWoodward37: Did you intend this class to be used for more than just date validation?

app/controllers/CalculationController.scala
```diff
@@ -190,6 +190,16 @@ trait CalculationController extends FrontendController {
     }
   }
 
+  val submitDisposalDate = Action { implicit request =>
+    disposalDateForm.bindFromRequest.fold(
+      errors => BadRequest(calculation.disposalDate(errors)),
+      success => {
+        calcConnector.saveFormData("acquisitionValue", success)
```
DavidWoodward37: Should this not be saved as disposalDate?

app/common/Validation.scala
MJCallahanPage: Potentially, I realised that the date validation would need to be in more than one place, which is why I tried to put this code in a 'shared/common' type of location. If we needed to add other functions to perform common validation they could go here. 

I'm not sure on what the 'best practice' is for things like this - may need some input from Mac.

app/controllers/CalculationController.scala
MJCallahanPage: Yep, I'll correct this now.

PR #102, opened 2016-04-20
Added validation to Allowable Losses.

jameshforster:


app/forms/AllowableLossesForm.scala
```diff
@@ -19,12 +19,37 @@ package forms
...
+      case "Yes" => {
+        data.allowableLossesAmt match {
+          case Some(data) => data >= 0
+          case None => true
+        }
+      }
+      case "No" => true
+    }
+  }
+
```
MJCallahanPage: Rather than repeat the same code here, you could just call the isPositive() function of the common.Validation object.

i.e.:
  ```
    case "Yes" => isPositive(data)
    case "No" => true
```

Can you also add a function to check for the numeric containing no more than two decimal places. I.e. is valid pounds and pence by calling `isMaxTwoDecimalPlaces()`

PR #104, opened 2016-04-20
Added validation for Other Reliefs.

jameshforster:


app/forms/OtherReliefsForm.scala
```diff
@@ -19,11 +19,20 @@ package forms
...
 
 object OtherReliefsForm {
+
+  def validateMinimum(data: BigDecimal): Boolean = {
+    data match {
+      case data if data < 0 => false
+      case _ => true
+    }
+  }
+
```
DavidWoodward37: I'd re-factor this to use the common methods that Matt introduced?

app/forms/OtherReliefsForm.scala
```diff
@@ -19,11 +19,20 @@ package forms
...
   val otherReliefsForm = Form(
     mapping(
-      "otherReliefs" -> bigDecimal
+      "otherReliefs" -> bigDecimal.verifying(Messages("calc.otherReliefs.errorMinimum"), otherReliefs => validateMinimum(otherReliefs))
```
DavidWoodward37: Same here

PR #126, opened 2016-04-29
Updated the calculate call methods on the Frontend to include time apportionment.

jameshforster:


test/connectors/CalculatorConnectorSpec.scala
```diff
@@ -65,6 +65,27 @@ class CalculatorConnectorSpec extends UnitSpec with MockitoSugar {
...
+    DisposalValueModel(150000),
+    AcquisitionCostsModel(None),
+    DisposalCostsModel(None),
+    EntrepreneursReliefModel("No"),
+    AllowableLossesModel("No", None),
+    CalculationElectionModel("time-apportioned-calculation"),
+    OtherReliefsModel(None),
+    OtherReliefsModel(None)
+  )
+
```
DavidWoodward37: Would it be worth moving this model into the CalculateRequestContructorSpec file you created that contain the other models?

test/constructors/CalculateRequestConstructorSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    "return a string from the flatCalcUrlExtra" in {
+      CalculateRequestConstructor.flatCalcUrlExtra(sumModel) shouldEqual "&improvementsAmt=0"
+    }
+
+    "return a string from the taCalcUrlExtra" in {
+      CalculateRequestConstructor.taCalcUrlExtra(sumModel) shouldEqual "&improvementsAmt=0&disposalDate=2010-10-10" +
+        "&acquisitionDate=1990-9-9"
+    }
```
DavidWoodward37: I get this tests that there is the correct extra part to the URL, but I think it will still be a flat-calculation type in the model?  I guess it's not that relevant but just a thought.

PR #128, opened 2016-04-29
CGT-879 created basic HTML and tests for acquisition date

miranda-hawkes:
Also fixed bug to set field to hidden in cgt.js rather than in HTML of formHiddenYesNoRadio helper. This is more user friendly for those with javascript not enabled.


jameshforster:
Please fix the merge conflicts with master.

PR #139, opened 2016-05-04
Updated summary page to work with time apportioned calc option.

jameshforster:


app/common/Dates.scala
```diff
@@ -22,19 +22,21 @@ import java.util.Date
...
-    }/$year")
+//    sf.parse(s"${day match {
+//      case day if day < 10 => "0" + day
+//      case _ => day
+//    }
+//    }/${month match {
+//      case month if month < 10 => "0" + month
+//      case _ => month
+//    }
+//    }/$year")
```
DavidWoodward37: Maybe remove these comments completely, if we need them we can always look back to a previous version :)

PR #146, opened 2016-05-04
Completed dynamic content and fixed coverage issues.

jameshforster:


DavidWoodward37:
Not sure what happened here man but I suspect you were running the back-end service at the same time as these are errors calling the localhost:9090/...

PR #157, opened 2016-05-06
CGT-931 added form binding and keystore for other reliefs rebased

miranda-hawkes:


macgrewal:
Update controller to get rebased calc result from connector (dummy implementation for now) and display the contents on the form.

PR #162, opened 2016-05-09
Some general cleanup and added functionality for the rebased calculation.

DavidWoodward37:


app/constructors/CalculationElectionConstructor.scala
```diff
@@ -36,27 +36,62 @@ trait CalculationElectionConstructor {
...
+            ("time", resultTime(summary, hc),
+              Messages("calc.calculationElection.message.time"),
+              Some(Messages("calc.calculationElection.message.timeDate")),
+              routes.CalculationController.otherReliefsTA().toString())
+          )
+        }
       }
       case "No" => {
-        Seq(("flat", resultFlat(summary, hc), Messages("calc.calculationElection.message.flat"), None, routes.CalculationController.otherReliefs().toString()))
+        Seq(("flat", resultFlat(summary, hc), Messages("calc.calculationElection.message.flat"),
```
jameshforster: No check in place for scenario where flat and rebased options are available but not time apportioned.

PR #163, opened 2016-05-09
CGT-933 added routing and post action for other reliefs rebased

miranda-hawkes:


jameshforster:
The view needs to be updated so the form is set to POST not GET.

PR #164, opened 2016-05-09
Updated Other Properties Model, Form and Page to include new drop dow…

jameshforster:
…n data capture for previous gains.


DavidWoodward37:
May need to fix the two new models that someone else has added!

DavidWoodward37:
Coverage on the Messages in the form is needed by travis.

PR #190, opened 2016-05-16
CGT-861 Amended Disposal Costs routing to PRR

miranda-hawkes:


MJCallahanPage:
Miranda, can you fix the merge conflicts please? Thanks

app/controllers/CalculationController.scala
```diff
@@ -380,7 +380,7 @@ trait CalculationController extends FrontendController {
       errors => Future.successful(BadRequest(calculation.disposalCosts(errors))),
       success => {
         calcConnector.saveFormData("disposalCosts", success)
-        Future.successful(Redirect(routes.CalculationController.entrepreneursRelief()))
+        Future.successful(Redirect(routes.CalculationController.privateResidenceRelief()))
```
MJCallahanPage: The routing is more complicated than this. Following Marks logic [here](https://confluence.tools.tax.service.gov.uk/display/CGT/Simple+PRR+Calculation).

There is a scenario where we can't do any PRR calc for them. That happens when they **don't** supply us an Acquisition Date **AND** they **don't** supply us a Rebased Value. 

If this happens we would jump straight to Entrpreneurs relief. 

If either of the dates **is** supplied then we can do some form of PRR calc so route to PRR.

app/controllers/CalculationController.scala
```diff
@@ -382,7 +382,22 @@ trait CalculationController extends FrontendController {
       errors => Future.successful(BadRequest(calculation.disposalCosts(errors))),
       success => {
         calcConnector.saveFormData(KeystoreKeys.disposalCosts, success)
-        Future.successful(Redirect(routes.CalculationController.entrepreneursRelief()))
+        calcConnector.fetchAndGetFormData[AcquisitionDateModel](KeystoreKeys.acquisitionDate).flatMap {
+          case Some(data) => data.hasAcquisitionDate match {
+            case "Yes" =>
+              Future.successful(Redirect(routes.CalculationController.privateResidenceRelief()))
+            case "No" => {
```
MJCallahanPage: What if it's not Yes or No. This would result in an inexhaustive case statement as there is no `case _` leg.

What might be better is:

```
case Some(data) if data.hasAcquisitionDate == "Yes" => match {
   Future.successful(Redirect(routes.CalculationController.privateResidenceRelief()))
case _ => {
    calcConnector.fetchAndGetFormData[RebasedValueModel](KeystoreKeys.rebasedValue).flatMap {
        case Some(rebasedData) if rebasedData.hasRebasedValue == "Yes" => {
           Future.successful(Redirect(routes.CalculationController.privateResidenceRelief()))
        }
        case _ => {
            Future.successful(Redirect(routes.CalculationController.entrepreneursRelief()))
         }
    }
```

PR #208, opened 2016-05-19
CGT-1041: Added dynamic back links based on user journey.

MJCallahanPage:
Apologies, this is a large PR - should have performed more commits.

Changed made to:

- Other Properties (Previous Taxable Gains/Disposals)
- Acquisition Date
- Improvements
- Entrepreneurs Relief
- Other reliefs
- Summary

Tests added for all permutations of user journey - including scenarios where the user jumps directly to pages and the back link can't be determined. In this scenario a new file has been introduced called DefaultRoutes with an entry "missingDataRoute" that links to the introduction page.

### DIFF
--- a/app/controllers/CalculationController.scala
+++ b/app/controllers/CalculationController.scala
@@ -43,7 +43,6 @@ import forms.RebasedCostsForm._
 import forms.PrivateResidenceReliefForm._
 import models._
 import java.util.{Calendar, Date}
-import play.api.Logger
 import play.api.mvc.{Result, AnyContent, Action}
 import uk.gov.hmrc.play.frontend.controller.FrontendController
 import uk.gov.hmrc.play.http.HeaderCarrier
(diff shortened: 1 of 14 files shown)


app/controllers/CalculationController.scala
```diff
@@ -210,21 +232,48 @@ trait CalculationController extends FrontendController {
   }
 
   //################### Acquisition Date methods #######################
+  def acquisitionDateBackUrl(implicit hc: HeaderCarrier): Future[String] = {
+    calcConnector.fetchAndGetFormData[OtherPropertiesModel](KeystoreKeys.otherProperties).map {
+      case Some(data) => data.otherPropertiesAmt match {
```
jameshforster: Needs to account for the yes/no question which overrides any input if no.

app/controllers/CalculationController.scala
```diff
@@ -210,21 +232,48 @@ trait CalculationController extends FrontendController {
...
+        case Some(value) if value > 0 => routes.CalculationController.annualExemptAmount().url
```
jameshforster: Should be when equal to 0

PR #213, opened 2016-05-20
Created common customer type keys

DavidWoodward37:


MJCallahanPage:
Can we add the use of these into the Test Specs as well. So that it future proofs the tests should we ever need to change the wording of the Customer Type.

PR #228, opened 2016-05-23
Updated gain in other reliefs pages to display with thousand separato…

jameshforster:
…rs and with no decimal places.

### DIFF
--- a/app/views/calculation/otherReliefsTA.scala.html
+++ b/app/views/calculation/otherReliefsTA.scala.html
@@ -2,6 +2,8 @@
 @import models._
 @import uk.gov.hmrc.play.views.html.{helpers => govHelpers}
 @import views.html.helpers._
+@import uk.gov.hmrc.play.views.helpers._
+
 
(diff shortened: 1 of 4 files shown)


app/views/calculation/otherReliefsTA.scala.html
```diff
@@ -19,9 +21,9 @@ <h1 class="heading-large">@Messages("calc.base.calculate.tax.heading")</h1>
 
         <div class="panel-indent">
             <ul class="list list-bullet">
-                <li id="totalGain">&pound;@SummaryConstructor.setPositive(result)<br><span class="font-xsmall">@SummaryConstructor.gainMessage(result)</span></li>
+                <li id="totalGain">&pound;@{MoneyPounds(result.totalGain,0).quantity}<br><span class="font-xsmall">@SummaryConstructor.gainMessage(result)</span></li>
```
MJCallahanPage: This has a setPositive before, so should this have .abs added after total gain so that it is always an absolute value.

e.g. `MoneyPounds(result.totalGain.abs,0).quantity`

PR #230, opened 2016-05-23
Will not work until James' fix for the js has been merged.

DavidWoodward37:


DavidWoodward37:
The scala tests will pass but acceptance will continue to fail until the fix from james for the js has been merged.

app/views/calculation/otherReliefs.scala.html
```diff
@@ -3,7 +3,7 @@
 @import uk.gov.hmrc.play.views.html.{helpers => govHelpers}
 @import views.html.helpers._
 
-@(otherReliefsForm: Form[OtherReliefsModel], result: CalculationResultModel, backUrl: String)(implicit request: Request[_])
+@(otherReliefsForm: Form[OtherReliefsModel], result: CalculationResultModel, buttonWording: Boolean, backUrl: String)(implicit request: Request[_])
```
MJCallahanPage: Can we change the Boolean name to be more meaningful? e.g. `hasExistingReliefAmount` or something like that so that it reads better in the if statement later.

i.e. `if(hasExistingReliefAmount) { ......`

Make it a little more obvious about what and why it's doing it

DavidWoodward37:
Acceptance test will now work with the latest master.

MJCallahanPage:
If they enter an amount of £0, what should happen? Currently it would show £0.00 on Calculation Election and then show 'Update Relief'. But should it actually have the text "Add relief" on Calculation Election and 'Add' on Other Reliefs?

One for Zed and Gregg maybe.

DavidWoodward37:
Hi Matt, can you have another look at this please?

PR #231, opened 2016-05-23
CGT-1082: Updated Calculation Election to match v2.6 of Prototype

MJCallahanPage:


app/views/calculation/calculationElection.scala.html
```diff
@@ -8,36 +8,56 @@
 
 @(calculationElectionForm: Form[CalculationElectionModel], summaryModel: SummaryModel, content: Seq[(String, String, String, Option[String], String)])(implicit request: Request[_], hc: HeaderCarrier)
 
-@main_template(Messages("calc.calculationElection.question"), articleLayout = false) {
+@sidebar = {
+    <ul>
+        <li>
+            <a id="helpLink1" href="https://www.gov.uk/guidance/capital-gains-tax-for-non-residents-calculating-taxable-gain-or-loss" target="_blank">
```
DavidWoodward37: This will need class="external-link" rel="external" attributes in the link

PR #233, opened 2016-05-23
Updated the personal allowance form validation and message

DavidWoodward37:
(approved by Greg).


jameshforster:
This is breaking the acceptance tests, could you look into the reasons why? It seems to occur when running a new scenario and fails to load a page.

PR #242, opened 2016-06-01
Added the overwrite messages and the placeholder text.

DavidWoodward37:


DavidWoodward37:
Will work with the latest master of the acceptance tests.

PR #244, opened 2016-06-02
Updated inputs to type number without spinner.

jameshforster:


MJCallahanPage:
If we introduce this change, we will re-introduce the bug in firefox where if a user enters a number with .000 then it re-renders the page without the decimal points and you can't submit past that page without re-entering the number entirely.

DavidWoodward37:
Team decision made to include this fix.  Will only break on rare cases on firefox as opposed to all cases on mobile devices.

PR #245, opened 2016-06-02
Updated PRR validation to make both fields mandatory when they are used.

jameshforster:


jameshforster:
Note: Merge with same branch in acceptance tests!

PR #250, opened 2016-06-02
Added invisible navigation text for external links.

jameshforster:


DavidWoodward37:
You'll need to include the hidden external link text for the link 
https://github.com/hmrc/capital-gains-calculator-frontend/blob/master/app/views/calculation/otherProperties.scala.html#L9-L11

and 

https://github.com/hmrc/capital-gains-calculator-frontend/blob/master/app/views/calculation/summary.scala.html#L50-L51

PR #264, opened 2016-06-13
CGT-1192 added the new acquisition costs template

DavidWoodward37:


app/views/helpers/formInputMoney.scala.html
```diff
@@ -12,6 +12,7 @@
         @moneyInputFormat(
             inputForm(fieldName),
             '_label -> questionText,
+            '_additionalContent -> Html(additionalHTMLContent.getOrElse("")),
```
MJCallahanPage: Because this is defaulted to space `""` it means that that the true false check in the helper will always return true.

This means there's a tiny grey box rendered. I don't think you need the getOrElse just handle the None in the helper.

app/views/helpers/moneyInputFormat.scala.html
```diff
@@ -25,6 +25,7 @@
 @value = @{ field.value match { case Some(x) => x case None => "" case x => x }}
 @labelAfter = @{ elements.args.get('_labelAfter).getOrElse(false).asInstanceOf[Boolean] }
 @labelHighlight = @{ elements.args.get('_labelHighlight).getOrElse(false).asInstanceOf[Boolean] }
+@additionalHTMLContent = @{ if(elements.args.contains('_additionalContent)) { true } else { false } }
```
MJCallahanPage: Currently, this always returns `true` based on the comment above.

conf/messages
```diff
@@ -115,6 +115,14 @@ calc.acquisitionDate.questionTwo = What was the date you signed the contract tha
 calc.acquisitionValue.question = How much did you pay for the property?
 calc.acquisitionValue.errorNegative = Enter a positive number for the amount you paid
 calc.acquisitionValue.errorDecimalPlaces = The amount you paid has too many numbers after the decimal point
+calc.acquisitionValue.bullet.title = Put the market value of the property instead if you:
+calc.acquisitionValue.bullet.one = inherited it
+calc.acquisitionValue.bullet.two = got it as a gift
+calc.acquisitionValue.bullet.three = bought it from a relative, business partner or
+calc.acquisitionValue.bullet.three.link = someone else you're connected to
```
MJCallahanPage: This needs a double single quote. `''` else the quote is ignored.

conf/messages
```diff
@@ -115,6 +115,14 @@ calc.acquisitionDate.questionTwo = What was the date you signed the contract tha
...
+calc.acquisitionValue.bullet.four = bought it for less than it’s worth because the seller wanted to help you
```
MJCallahanPage: This should also be double single quote

PR #269, opened 2016-06-13
CGT-1227: Removed need for start page by moving Keystore to customer-…

MJCallahanPage:
…type.

Acceptance Test changes on same branch name need to be merged at same time.


app/controllers/CalculationController.scala
```diff
@@ -69,9 +72,15 @@ trait CalculationController extends FrontendController {
 
   //################### Customer Type methods #######################
   val customerType = Action.async { implicit request =>
-    calcConnector.fetchAndGetFormData[CustomerTypeModel](KeystoreKeys.customerType).map {
-      case Some(data) => Ok(calculation.customerType(customerTypeForm.fill(data)))
-      case None => Ok(calculation.customerType(customerTypeForm))
+    if (request.session.get(SessionKeys.sessionId).isEmpty) {
```
DavidWoodward37: This is more a note for myself, but I think this check is going to be invalidated by the predicate I set up for the session timeout as it is basically the same check.  I think I am going to need to remove the check from the customer start page.  Otherwise no user will ever be able to actually enter the calculator.

DavidWoodward37:
The only other point was are there any changes to the acceptance tests done for this yet?

PR #271, opened 2016-06-14
CGT-1189 Added Yes/No option to add a relief if Flat is the only calc option 

miranda-hawkes:
- This included creating a new view **otherReliefsFlat** that is shown when page is opened from calculationElection for consistency with otherReliefsRebased and otherReliefsTa
- **otherReliefs** view is used only for Flat calc journey where no other calc types are possible and now incorporates a Yes/No dropdown where users are asked if they would like to add a relief
- Summary page now caters for the above if the user selects 'No'
- Where total gain and taxable gain were shown as bullet points on other reliefs pages, this has been changed reflecting the latest prototype
- formHiddenYesNoRadio helper amended to decrease a large empty gap on other reliefs where no second question is asked in the hidden drop down


miranda-hawkes:
Only merge with CGT-1204 acceptance tests

MJCallahanPage:
Hi Miranda, can you merge latest master please. Thanks

PR #272, opened 2016-06-15
Added FeatureLock trait and class.

jameshforster:


conf/application.conf
```diff
@@ -38,6 +38,9 @@ Dev {
       }
     }
   }
+  features {
+    RTT = true
```
MJCallahanPage: Shouldn't we have this set to false?

test/controllers/FeedbackControllerSpec.scala
```diff
@@ -71,6 +71,7 @@ class FeedbackControllerSpec extends UnitSpec with MockitoSugar with WithFakeApp
       override val analyticsToken: String = ""
       override val reportAProblemPartialUrl: String = ""
       override val contactFormServiceIdentifier: String = ""
+      override val featureRTTEnabled = true
```
MJCallahanPage: Shouldn't this be default to false?

MJCallahanPage:
This needs to be updated to include unit tests for the new predicate controller.

MJCallahanPage:
Added tests for the FeatureLock implementation.

@DavidWoodward37 - can you review this PR please as both James and I have worked on it. Thanks.

PR #276, opened 2016-06-16
Version control refactor. Please read the comments.

DavidWoodward37:
<h1>Comments!</h1>

The common folder has a new package called nonresident: 

<ul>
 <li>Calc and customer type have been moved into it as they are not applicable for RTT.</li>

<li> Keystore keys was also added as for now at least having identical keys for both calculators would cause strange issues if a user ever jumped between the two (not sure if this will happen yet but...).  Doesn't necessarily need to be in the sub-directory the keys could just have a different structure but it's there for now. </li>
</ul>

The calculator connector has been moved to a separate nonresidents package as well, @MJCallahanPage had a suggestion to refactor it similar to how each constructor deals with only one thing for instance a separate connector for the constants and the calculations. 

The controllers for the calculation have been refactored into a nonresident package too with timeout and feedback being left outside.

Final change is all models and forms have been moved as the prototype for RTT looks to be using a different style to capture data so splitting them would be ideal, it also allows us to customise the error messages per flow (maybe not for consistency but the option is there).

One final thing, we may need to put in a request to the business (or whomever manages gov.uk) to get the link updated on the page to direct people to the calculator as the route now has /non-residents/ in it.


DavidWoodward37:
There is also a need for a change to the acceptance tests as they verify by their URL which has now changed. And a potential refactor to the back end as well to reflect the change.

macgrewal:
All values for common.nonresident.KeyStoreKeys need prefixing with 'nr:' so that there is no conflict with their resident counterparts

macgrewal:
Undo refactor on connectors.nonresident.CalculatorConnector (back to connectors.CalculatorConnector). A connector represents a means of connecting to another microservice not different functionality within the same microservice.

PR #291, opened 2016-06-24
CGT-1306: Added for, model, html, post route, post action for Disposal Costs

MJCallahanPage:


macgrewal:
Please fix broken unit tests. It looks like you're not mocking out the connector properly (keystore calls)

app/forms/resident/DisposalCostsForm.scala
```diff
@@ -0,0 +1,43 @@
...
+import play.api.i18n.Messages
+import common.Validation._
+import uk.gov.hmrc.play.views.helpers.MoneyPounds
+
+object DisposalCostsForm {
+
+  val disposalCostsForm = Form(
+    mapping(
+      "amount" -> text
+        .verifying(Messages("calc.base.undefinedMessage"), amt => amt != "")
```
miranda-hawkes: You could call isNotEmpty instead of using != "" here

app/forms/resident/DisposalCostsForm.scala
```diff
@@ -0,0 +1,43 @@
...
+        .verifying(Messages("calc.base.undefinedMessage"), amt => isBigDecimalNumber(amt))
+        .transform[BigDecimal](amt => BigDecimal(amt), amt => amt.toString())
+        .verifying(Messages("calc.common.error.maxNumericExceeded") +
+          MoneyPounds(Constants.maxNumeric, 0).quantity + " " +
+          Messages("calc.common.error.maxNumericExceeded.OrLess"),
+          amt => isLessThanEqualMaxNumeric(amt))
+        .verifying(Messages("calc.base.undefinedMessage"), amt => isGreaterThanZero(amt))
```
miranda-hawkes: We haven't got business validation yet, but I think its likely disposal costs will go the same way as in NRCGT - therefore we should probably be calling isPositive (which allows zero) here. We can always change it later

app/controllers/resident/GainController.scala
```diff
@@ -97,9 +98,22 @@ trait GainController extends FeatureLock {
   val acquisitionCosts = FeatureLockForRTT.async { implicit request =>
     Future.successful(Ok(views.acquisitionCosts()))
   }
-  
+
+  //################# Disposal Costs Actions ########################
```
miranda-hawkes: Might be just being picky... but can we put the Disposal Costs actions directly after Disposal Value and before Acquisition Value so it matches the user journey?

test/forms/resident/DisposalCostsFormSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+
+package forms.resident
+
+import uk.gov.hmrc.play.test.{UnitSpec, WithFakeApplication}
+import models.resident.DisposalCostsModel
+import forms.resident.DisposalCostsForm._
+import assets.{MessageLookup => commonMessages}
+import common.Constants
+import controllers.helpers.FakeRequestHelper
+import play.api.i18n.Messages
```
miranda-hawkes: Remove unused import and play.api.mvc.Request

PR #298, opened 2016-06-29
CGT-1307 - Added investments view content and validation

macgrewal:


test/common/ValidationSpec.scala
```diff
@@ -120,4 +120,109 @@ class ValidationSpec extends UnitSpec {
...
-}
+
+  "calling bigDecimalCheck" when {
+
+    "input contains non-numeric characters" should {
+      "fail" in {
+        bigDecimalCheck("abc") shouldBe false
+      }
+    }
+
```
DavidWoodward37: Do you want a validation check to also check for the "" => true condition.

test/forms/resident/ImprovementsFormSpec.scala
```diff
@@ -0,0 +1,196 @@
...
+        form.value.get shouldBe ImprovementsModel(BigDecimal(1))
+      }
+    }
+
+    "supplied with valid amount" should {
+      "not raise form error" in {
+        val form = improvementsForm.bind(Map("amount" -> "1"))
+        form.hasErrors shouldBe false
+      }
+    }
```
DavidWoodward37: Could you refactor the two above tests into the same "supplied with a valid amount" should "...." in {.....} "....." in {....} then you could also use the val of the form for both?

test/assets/MessageLookup.scala
```diff
@@ -82,8 +82,11 @@ object MessageLookup {
   }
 
   //Improvements messages
-  object improvements {
+  object improvementsView {
```
DavidWoodward37: Should we change the other objects to this naming style too?

macgrewal: changed to stop scala-style whingeing. We could do this on other views also.

PR #312, opened 2016-06-30
Added HTML content and submit action for other properties

miranda-hawkes:
- Also refactored reliefs and other properties to use Boolean instead of String in form and model
- amended reliefs html to correctly space button and error line
- added dynamic back link


jameshforster:
Please transform the boolean back to a Yes/No value instead of using a toString
(see function in transformers package)

PR #313, opened 2016-06-30
Added model, form, HTML, POST and keystore for Brought Forward Losses Value

jameshforster:


jameshforster:
Trying to fix travis builds.

PR #324, opened 2016-07-01
Updated Improvements route to go to Reliefs if a positive gain is acq…

jameshforster:
…uired.


test/controllers/resident/GainControllerTests/ImprovementsActionSpec.scala
```diff
@@ -17,11 +17,12 @@
 package controllers.resident.GainControllerTests
 
 import assets.MessageLookup.{improvementsView => messages}
+import common.Dates._
```
miranda-hawkes: This doesn't seem to be used - i'll remove it

PR #341, opened 2016-07-05
Fixed broken post action from Previous Taxable Gains.

jameshforster:


miranda-hawkes:
Can you add a test for a valid form submit please?

PR #342, opened 2016-07-05
Updated routing from AEA/brought forward losses/brought forward losses value

miranda-hawkes:
Updated so:
- negative and zero chargeable gain routes to summary
- entry of zero on AEA page and a positive chargeable gain routes to previous-taxable-gains
- other positive gains route to current-income unless other screens before are required
- also added spacing in AEA action spec as it was hard to read before

### DIFF
--- a/app/controllers/resident/DeductionsController.scala
+++ b/app/controllers/resident/DeductionsController.scala
@@ -71,11 +71,13 @@ trait DeductionsController extends FeatureLock {
     }
 
     reliefsForm.bindFromRequest.fold(
-      errors => {for {
-        answerSummary <- answerSummary(hc)
-        totalGain <- totalGain(answerSummary, hc)
-        route <- routeRequest(errors, totalGain)
-      } yield route},
+      errors => {
+        for {
+          answerSummary <- answerSummary(hc)
+          totalGain <- totalGain(answerSummary, hc)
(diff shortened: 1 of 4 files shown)


app/controllers/resident/DeductionsController.scala
```diff
@@ -258,7 +258,18 @@ trait DeductionsController extends FeatureLock {
         errors => Future.successful(BadRequest(views.lossesBroughtForward(errors, backUrl))),
         success => {
           calcConnector.saveFormData[LossesBroughtForwardModel](KeystoreKeys.ResidentKeys.lossesBroughtForward, success)
-          broughtForwardLossesContinueRoute(success)
+          for {
+            broughtForwardLossesClaimed <- claimingBFLossesCheck(success)
+            otherPropertiesClaimed <- otherPropertiesCheck
+            positiveChargeableGain <- positiveChargeableGainCheck
```
jameshforster: Cannot be called here, breaks if there is no value for Brought Forward Losses Value. Calculation check should only be performed if the "No" option is selected.

jameshforster: Again, if it would normally route to the AEA page you cannot call the calculation as the value may not yet be given and therefore breaks on the getChargeableGainAnswers method when trying to fetch the non-existent AEA.

PR #346, opened 2016-07-07
Added validation for no decimal places for Personal Allowance.

jameshforster:


morgan-fernandes:
Need to change the test view where we accepting 3 decimal places Currently we still keep getting the message to enter a value to the nearest decimal

jameshforster:
Updated the view to match the validation changes.

PR #359, opened 2016-07-11
Added fix so only used values are displayed in deductions on Final Su…

jameshforster:
…mmary.


DavidWoodward37:
Might be worth writing one or two tests to check if 0 is displayed despite values being supplied?

PR #386, opened 2016-07-18
CGT-1470: Added validation for four digit date year input

MJCallahanPage:


macgrewal:
The date 1/1/1900 fails validation. Is this expected? If the intent is to make the year 4 digits then this validation is not complete.

MJCallahanPage:
I spoke to James about this, I originally had it just check for 4 digits, so the user could enter a year back from 1000 to 9999.

James said he'd had a convo with you about only accepting years greater than 1900. As that would be more than earlier enough?

Happy to go with whichever approach?

MJCallahanPage:
Capital Gains Tax was first introduced in 1965, so there's also an argument that the business might decide that this is the earliest disposal date they can enter? Not sure. 

Although, if they trying to declare something from that long ago our service isn't really applicable haha!

PR #394, opened 2016-07-19
CGT-1485: Session Timeout now dynamic based on user journey

MJCallahanPage:
**Change Description**
To implement this change the FeatureLock and ValidActiveSession predicate traits have both been amended. 

FeatureLock has a val featureEnabled which is only set to true from its default position of false if the ApplicationConfig is set to true. 

This is achieved by overriding the val in the FeatureLockForRTT Object which extends the trait. This removed the need for an internal class called FeatureLockFor. The removal of this was required so that the Object could also override the val sessionTimeoutUrl in ValidActiveSession trait - see below.

There is now an additional val in the ValidActitveSession trait called sessionTimourUrl which is overridden by the Object which extends the trait. This allows each FeatureLock object to override with a specific journey start URL.

For the existing non-resident code the controller overrides the sessionTimeoutUrl as we have not implemented any feature locks within that controller.

This implementation hopefully allows for simplified scaling for further features which have different journey start points (urls).

**Note**
Moved the session-timeout route to be "calculate-your-capital-gains/session-timeout" rather than "calculate-your-capital-gains/non-resident/session-timeout"

This initially caused issues and required both the Assets routes and Feedback routes to be moved to app.Routes as well.

### DIFF
--- a/app/controllers/predicates/FeatureLock.scala
+++ b/app/controllers/predicates/FeatureLock.scala
@@ -17,42 +17,41 @@
 package controllers.predicates
 
 import config.ApplicationConfig
-import controllers.routes
 import play.api.mvc.{Action, Result, AnyContent, Request}
 
 import scala.concurrent.Future
 
 trait FeatureLock extends ValidActiveSession {
 
+  val featureEnabled: Boolean = false
   private type PlayRequest = Request[AnyContent] => Result
   private type AsyncPlayRequest = Request[AnyContent] => Future[Result]
 
-  class FeatureLockFor(condition: Boolean) {
-
-    def async(action: AsyncPlayRequest): Action[AnyContent] = {
-      ValidateSession.async { implicit request =>
-        if (condition) {
-          action(request)
-        }
-        else {
-          Future.successful(NotFound)
-        }
+  def async(action: AsyncPlayRequest): Action[AnyContent] = {
+    ValidateSession.async { implicit request =>
+      if (featureEnabled) {
+        action(request)
       }
-    }
-
-    def asyncNoTimeout(action: AsyncPlayRequest): Action[AnyContent] = {
-      Action.async { implicit request =>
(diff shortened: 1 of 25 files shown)


app/controllers/predicates/FeatureLock.scala
```diff
@@ -17,42 +17,41 @@
...
 
 import config.ApplicationConfig
-import controllers.routes
 import play.api.mvc.{Action, Result, AnyContent, Request}
 
 import scala.concurrent.Future
 
 trait FeatureLock extends ValidActiveSession {
 
+  val featureEnabled: Boolean = false
```
DavidWoodward37: See comment below

app/controllers/predicates/FeatureLock.scala
```diff
@@ -17,42 +17,41 @@
...
       }
     }
   }
 
-  object FeatureLockForRTT extends FeatureLockFor(RTTCondition)
-
-  lazy val RTTCondition = ApplicationConfig.featureRTTEnabled
-}
+  object FeatureLockForRTT extends FeatureLock {
+    lazy val RTTEnabled = ApplicationConfig.featureRTTEnabled
```
DavidWoodward37: You can remove this line entirely and move the ApplicationConfig down to the next line if you make the featureEnabled value a lazy one.  It just saves single use of a val and also as featureEnabled isn't a lazy val RTTEnabled is actually called as soon as you create a FeatureLock, sort of removing the lazy nature of it.

app/controllers/predicates/FeatureLock.scala
```diff
@@ -17,42 +17,41 @@
...
+    override val featureEnabled = RTTEnabled
```
DavidWoodward37: e.g. override lazy val featureEnabled = ApplicationConfig.featureRTTEnabled

PR #408, opened 2016-07-21
CGT-1495: Copied outsideTaxYear controller action, routes and tests

MJCallahanPage:
**Important**

All the existing shared views have been amended to use the new template - this meant that a homeLink parameter needs to be passed to them. For now, there is a string **"TODO - Needs Parameter"** - this will have to be replaced when future tasks are worked on!

**Notes**

- re-uses the existing view, which has been amended to accept two new parameters for homeLink and continue action.

- Includes 'Home' link changes for error pages - this is dynamically determined based on the URL path in the frontendGlobal config. New test class added for this.

- Includes new template resident_main_template to pass the home link based on the journey to the govUkWrapper. (replaces the previous template)

- New resident_properties_main_template that sets the homelink to be the start of the resident_properties journey.

- The existing main_template (used by NRCGT) has been changed to default to the start of the NRCGT route

### DIFF
--- a/app/common/KeystoreKeys.scala
+++ b/app/common/KeystoreKeys.scala
@@ -62,4 +62,8 @@ object KeystoreKeys {
     val personalAllowance = "res:property:personalAllowance"
   }
 
+  object ResidentSharesKeys {
+    val disposalDate = "res:shares:disposalDate"
+  }
+
 }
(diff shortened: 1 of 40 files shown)


DavidWoodward37:
Sorry, please fix the merge conflicts!

PR #411, opened 2016-07-22
Added disposal value for shares page

miranda-hawkes:
- Fixed a tiny bug in helper that created empty form-hint tags when not necessary which could be confusing for accessibility reasons
- Refactored package names for some property action specs

### DIFF
--- a/app/controllers/resident/shares/GainController.scala
+++ b/app/controllers/resident/shares/GainController.scala
@@ -38,16 +38,18 @@ trait GainController extends FeatureLock {
 
   val calcConnector: CalculatorConnector
 
+  private val homeLink = controllers.resident.shares.routes.GainController.disposalDate().url
+
   //################# Disposal Date Actions ####################
   val disposalDate = FeatureLockForRTTShares.asyncNoTimeout { implicit request =>
     if (request.session.get(SessionKeys.sessionId).isEmpty) {
       val sessionId = UUID.randomUUID.toString
-      Future.successful(Ok(views.disposalDate(disposalDateForm)).withSession(request.session + (SessionKeys.sessionId -> s"session-$sessionId")))
+      Future.successful(Ok(views.disposalDate(disposalDateForm, homeLink)).withSession(request.session + (SessionKeys.sessionId -> s"session-$sessionId")))
     }
     else {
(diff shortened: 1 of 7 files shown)


DavidWoodward37:
Please resolve the merge conflicts! :)

app/controllers/resident/shares/GainController.scala
```diff
@@ -86,6 +87,24 @@ trait GainController extends FeatureLock {
   }
 
   //################ Disposal Value Actions ######################
-  val disposalValue = TODO
+  val disposalValue = FeatureLockForRTT.async { implicit request =>
+    calcConnector.fetchAndGetFormData[DisposalValueModel](keystoreKeys.disposalValue).map {
+      case Some(data) => Ok(views.disposalValue(disposalValueForm.fill(data)))
+      case None => Ok(views.disposalValue(disposalValueForm))
```
DavidWoodward37: There's a couple of things you could do here.

- Matt has a new way of doing the fetch for Keystore data if you look at his CGT-1499 pull request.  Maybe changing if for some consistency might be good? 

- You'll need to implement a different home link (Again see CGT-1499 for an example).

- It might worth defining that home link as a private val at the top of the class to .../shares/disposal-date.  The it can be re-used in every view.  It won't save much code redundancy in this PR specifically but it will for others.

app/views/calculation/resident/shares/gain/disposalValue.scala.html
```diff
@@ -0,0 +1,26 @@
+@import models.resident.DisposalValueModel
+@import uk.gov.hmrc.play.views.html.helpers.form
+@import views.html.helpers._
+@import views.html.calculation.resident.shares._
+
+@(disposalValueForm : Form[DisposalValueModel])(implicit request: Request[_])
```
DavidWoodward37: Home link needs to be passed in here.

test/controllers/resident/properties/DeductionsControllerSpec/AllowableLossesValueActionSpec.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package controllers.resident.properties.DeductionsControllerTests
+package controllers.resident.properties.DeductionsControllerSpec
```
DavidWoodward37: Thank you for changing these.

PR #413, opened 2016-07-22
Added the new Allowable Losses controller and specs.

DavidWoodward37:


app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -218,8 +218,14 @@ trait DeductionsController extends FeatureLock {
...
-        case None => Ok(commonViews.allowableLossesValue(allowableLossesValueForm, taxYear))
+        case Some(data) => Ok(commonViews.allowableLossesValue(allowableLossesValueForm.fill(data), taxYear,
+          controllers.resident.properties.routes.GainController.disposalDate().toString,
+          controllers.resident.properties.routes.DeductionsController.submitAllowableLossesValue(),
+          Some(controllers.resident.properties.routes.DeductionsController.allowableLosses().toString)))
+        case None => Ok(commonViews.allowableLossesValue(allowableLossesValueForm, taxYear,
+          controllers.resident.properties.routes.GainController.disposalDate().toString,
+          controllers.resident.properties.routes.DeductionsController.submitAllowableLossesValue(),
+          Some(controllers.resident.properties.routes.DeductionsController.allowableLosses().toString)))
       }
```
MJCallahanPage: The "Ok(commonViews.allowable..." could be refactored out. The map is peformed to determine whether to populate the form with data.

This could be done in a seperate function called getKeystoreData(), call that in the for comprehension and then simplify oute request to just be the "Ok(......"

If you look at my PR for CGT-1499 it has an example of this for lossesBroughtForwardValue.

Not necessarily saying it's any better but it's just a different option with less duplication.

app/controllers/resident/shares/DeductionsController.scala
```diff
@@ -109,7 +109,53 @@ trait DeductionsController extends FeatureLock {
 
   //################# Allowable Losses Value Actions ############################
 
-  val allowableLossesValue = TODO
+  private val allowableLossesHomeLink = controllers.resident.shares.routes.GainController.disposalDate().toString
+  private val allowableLossesPostAction = controllers.resident.shares.routes.DeductionsController.submitAllowableLossesValue()
+  private val allowableLossesBackLink = Some(controllers.resident.shares.routes.DeductionsController.allowableLosses().toString)
+
```
MJCallahanPage: I like this. I should have done this in my PR. Could one to roll out to the team?

app/controllers/resident/shares/DeductionsController.scala
```diff
@@ -109,7 +109,53 @@ trait DeductionsController extends FeatureLock {
...
+      calcConnector.fetchAndGetFormData[AllowableLossesValueModel](keystoreKeys.allowableLossesValue).map {
+        case Some(data) => Ok(commonViews.allowableLossesValue(allowableLossesValueForm.fill(data), taxYear,
+          allowableLossesHomeLink,
+          allowableLossesPostAction,
+          allowableLossesBackLink))
+        case None => Ok(commonViews.allowableLossesValue(allowableLossesValueForm, taxYear,
+          allowableLossesHomeLink,
+          allowableLossesPostAction,
+          allowableLossesBackLink))
+      }
```
MJCallahanPage: Again, you could do the same refactor retrieve the keystore data and bind it to a form prior to the route request to avoid code duplication.

PR #415, opened 2016-07-22
Copied AEA page for shares.

jameshforster:


app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -404,10 +404,27 @@ trait DeductionsController extends FeatureLock {
 
   //################# Annual Exempt Amount Input Actions #############################
   val annualExemptAmount = FeatureLockForRTT.async { implicit request =>
-    calcConnector.fetchAndGetFormData[AnnualExemptAmountModel](keystoreKeys.annualExemptAmount).map {
-      case Some(data) => Ok(commonViews.annualExemptAmount(annualExemptAmountForm().fill(data)))
-      case None => Ok(commonViews.annualExemptAmount(annualExemptAmountForm()))
+
+    val postAction = controllers.resident.properties.routes.DeductionsController.submitAnnualExemptAmount
```
DavidWoodward37: For now I think you can move this outside of the val into a private val (See CGT-1497 if you want an example). Just to as this route will be the same as the form with errors on the POST.

app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -404,10 +404,27 @@ trait DeductionsController extends FeatureLock {
...
+
+    def routeRequest(backLink: Option[String]) = {
+      calcConnector.fetchAndGetFormData[AnnualExemptAmountModel](keystoreKeys.annualExemptAmount).map {
+        case Some(data) => Ok(commonViews.annualExemptAmount(annualExemptAmountForm().fill(data), backLink, postAction))
+        case None => Ok(commonViews.annualExemptAmount(annualExemptAmountForm(), backLink, postAction))
```
DavidWoodward37: Maybe refactor this out as Matt did? I'm not sure if it makes it better or worse.  It gets us closer to the one function does one and only one thing I guess.

app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -428,9 +454,9 @@ trait DeductionsController extends FeatureLock {
       calcConnector.getFullAEA(taxYear)
     }
 
-    def routeRequest(maxAEA: BigDecimal): Future[Result] = {
+    def routeRequest(maxAEA: BigDecimal, backLink: Option[String]): Future[Result] = {
       annualExemptAmountForm(maxAEA).bindFromRequest.fold(
-        errors => Future.successful(BadRequest(commonViews.annualExemptAmount(errors))),
+        errors => Future.successful(BadRequest(commonViews.annualExemptAmount(errors, backLink, postAction))),
```
DavidWoodward37: You'll need to add the new home link change too.  (See one of matt's PR's.)

test/views/resident/AnnualExemptAmountViewSpec.scala
```diff
@@ -120,20 +122,37 @@ class AnnualExemptAmountViewSpec extends UnitSpec with WithFakeApplication with
...
+      "have the correct back link text" in {
+        backLink.text shouldBe commonMessages.calcBaseBack
+      }
+      "have the back-link class" in {
+        backLink.hasClass("back-link") shouldBe true
+      }
+      "have a link to Brought Forward Losses Value" in {
+        backLink.attr("href") shouldBe controllers.resident.shares.routes.DeductionsController.lossesBroughtForwardValue().toString
+      }
+    }
```
DavidWoodward37: You'll need to update here with a home link test!

DavidWoodward37:
And some acceptance tests too please!

PR #419, opened 2016-07-25
Created view, action and route for resident shares Current Income.

jameshforster:


app/views/calculation/resident/shares/income/currentIncome.scala.html
```diff
@@ -0,0 +1,33 @@
...
+
+    @errorSummary(currentIncomeForm, "current-income")
+
+
+    <h1 class="heading-large">
+        @if(taxYear.taxYearSupplied == "2016/17"){@Messages("calc.resident.currentIncome.questionCurrentYear")
+        } else {@Messages("calc.resident.currentIncome.question", taxYear.taxYearSupplied)}
+    </h1>
+
+    @form(action = controllers.resident.shares.routes.IncomeController.submitCurrentIncome){
```
morgan-fernandes: Is there a reason why we cannot use the postAction?

PR #423, opened 2016-07-26
Added GET action for first summary

miranda-hawkes:
- Also added part of html view to be able to call from controller

### DIFF
--- a/test/controllers/resident/shares/SummaryActionSpec.scala
+++ b/test/controllers/resident/shares/SummaryActionSpec.scala
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2016 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


test/controllers/resident/shares/SummaryActionSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    when(mockCalculatorConnector.getTaxYear(Matchers.any())(Matchers.any()))
+      .thenReturn(Future.successful(taxYearModel))
+
+    new SummaryController {
+      override val calculatorConnector: CalculatorConnector = mockCalculatorConnector
+    }
+  }
+
+  "Calling .summary from the SummaryController" when {
```
DavidWoodward37: Maybe add Shares in here to make it easier to identify if a test does go wrong? 

- "Calling .summary from the shares SummaryController"

PR #424, opened 2016-07-26
Fixed use of magic numbers and now calls AEA from backend.

jameshforster:


miranda-hawkes:
Sorry James I should've merged this before doing the shares Summary Controller which you just merged, would you be able to put these changes into shares summary controller too as part of this please?

PR #426, opened 2016-07-27
Added shares Gain Summary view.

DavidWoodward37:


app/controllers/resident/shares/SummaryController.scala
```diff
@@ -35,7 +35,7 @@ trait SummaryController extends FeatureLock {
 
   val calculatorConnector: CalculatorConnector
 
-  private val homeLink = controllers.resident.shares.routes.GainController.disposalDate().url
+//  private val homeLink = controllers.resident.shares.routes.GainController.disposalDate().url
```
miranda-hawkes: I think the homeLink should still be passed into the view the same as the other views - then when it changes we can just update this private val on each controller instead of every screen

app/views/calculation/resident/shares/summary/gainSummary.scala.html
```diff
@@ -1,14 +1,81 @@
...
+    @if(dateInsideAcceptedTaxYears(getDay(answers.disposalDate), getMonth(answers.disposalDate), getYear(answers.disposalDate))) {
+
+        @if(gain < 0) {
+            <section id="whatToDoNext" class="summary-underline summary-overline">
+                <h3 id="whatToDoNextTitle" class="bold-medium">@Messages("calc.resident.summary.whatNext")</h3>
+                <div id="whatToDoNextText" class="font-medium what-to-do-next-text">@Messages("calc.resident.summary.whatNext.youCan") <a id="whatToDoNextLink"
+                                                                                               class="external-link"
+                                                                                               rel="external"
+                                                                                               href="https://www.gov.uk/tax-sell-property/what-you-pay-it-on"
+                                                                                               target="_blank"
```
miranda-hawkes: Link needs to be to https://www.gov.uk/capital-gains-tax/losses as per prototype

test/views/resident/shares/summary/GainSummaryViewSpec.scala
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    "have a link" which {
+
+      "should have a href attribute" in {
+        doc.select("#whatToDoNextLink").hasAttr("href") shouldEqual true
+      }
+
+      "should link to the what-you-pay-on-it govuk page" in {
+        doc.select("#whatToDoNextLink").attr("href") shouldEqual "https://www.gov.uk/tax-sell-property/what-you-pay-it-on"
```
miranda-hawkes: This'll also need to change

test/views/resident/shares/summary/GainSummaryViewSpec.scala
```diff
@@ -0,0 +1,356 @@
...
+import org.jsoup.Jsoup
+import uk.gov.hmrc.play.test.{UnitSpec, WithFakeApplication}
+import views.html.calculation.resident.shares.{summary => views}
+
+class GainSummaryViewSpec extends UnitSpec with WithFakeApplication with FakeRequestHelper {
+
+  "Summary view" should {
+
+    val testModel = GainAnswersModel(
+      constructDate(12, 9, 1990),
```
miranda-hawkes: This might be picky, but 12/9/1990 isn't in the 2015/16 tax year as per the lazy val taxYearModel :P

test/views/resident/shares/summary/GainSummaryViewSpec.scala
```diff
@@ -0,0 +1,356 @@
...
+
+      lazy val backLink = doc.getElementById("back-link")
+
+      "has the id 'back-link'" in {
+        backLink.attr("id") shouldBe "back-link"
+      }
+
+      s"has the text '${commonMessages.calcBaseBack}'" in {
+        backLink.text shouldBe commonMessages.calcBaseBack
+      }
```
miranda-hawkes: Could you add a test for the back link url please? As it isn't being checked in the controller because it is fixed

PR #439, opened 2016-07-28
Added conditional content to brought forward losses and update messag…

jameshforster:
…e for residents.

### DIFF
--- a/test/views/resident/LossesBroughtForwardViewSpec.scala
+++ b/test/views/resident/LossesBroughtForwardViewSpec.scala
@@ -46,8 +46,11 @@ class LossesBroughtForwardViewSpec extends UnitSpec with WithFakeApplication wit
       doc.select("fieldset").attr("aria-details") shouldBe "help"
     }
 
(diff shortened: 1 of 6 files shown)


test/views/resident/LossesBroughtForwardViewSpec.scala
```diff
@@ -42,6 +42,40 @@ class LossesBroughtForwardViewSpec extends UnitSpec with WithFakeApplication wit
       doc.title() shouldBe messages.title("2015/16")
     }
 
+    "have a fieldset with aria-details attribute" in {
+      doc.select("fieldset").attr("aria-details") shouldBe "help"
+    }
+
+    s"have a drop down button with the text ${messages.helpInfoTitle}" in {
+      doc.body.getElementsByTag("summary").attr("role") shouldBe "button"
+      doc.body.getElementsByTag("summary").text shouldEqual messages.helpInfoTitle
```
DavidWoodward37: I think you should split this into two tests! Only one assert per test!

PR #441, opened 2016-07-28
Updated what to do next for zero gain or loss.

jameshforster:


DavidWoodward37:
Update to the final view summary's message too! (properties -> shares).

PR #449, opened 2016-08-03
Added form and model for PRR value

miranda-hawkes:


test/forms/resident/PrivateResidenceReliefValueFormSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+    "supplied with no data for amount" should {
+      lazy val form = privateResidenceReliefValueForm.bind(Map(("amount", "")))
+
+      "return a form with errors" in {
+        form.hasErrors shouldBe true
+      }
+
+      s"return a form with the error message '${errorMessages.mandatoryAmount}'" in {
+        form.error("amount").get.message shouldBe errorMessages.mandatoryAmount
+      }
```
DavidWoodward37: I think we may also need to add in a test for the length of the error array e.g.

"raise 1 form error" in {
        form.errors.length shouldBe 1
      }

PR #451, opened 2016-08-03
Added reliefs used value summary and report from backend.

jameshforster:
Merge with CGT-1635


miranda-hawkes:
On the summaries it should say 'Other tax reliefs used' instead of 'Reliefs'

PR #453, opened 2016-08-04
Updated resident properties reliefs value page with new PRR content.

jameshforster:


app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -102,15 +102,27 @@ trait DeductionsController extends FeatureLock {
...
+    } yield route
   }
 
   val submitReliefsValue = FeatureLockForRTT.async { implicit request =>
     reliefsValueForm.bindFromRequest.fold(
-      errors => Future.successful(BadRequest(views.reliefsValue(errors, homeLink))),
+      errors => for {
+        answerSummary <- answerSummary(hc)
+        totalGain <- totalGain(answerSummary, hc)
+      } yield BadRequest(views.reliefsValue(errors, homeLink, totalGain)),
```
DavidWoodward37: So out of curiosity why the nested for-yeild here rather than say the structure we had for instance here?

https://github.com/hmrc/capital-gains-calculator-frontend/blob/master/app/controllers/resident/shares/DeductionsController.scala#L104-L125

I think we should probably gather the rest of the dev team around to discuss maybe?

PR #456, opened 2016-08-04
Added GET action and keystore to PRR value

miranda-hawkes:
It made sense to implement keystore as part of this

### DIFF
--- a/app/controllers/resident/properties/DeductionsController.scala
+++ b/app/controllers/resident/properties/DeductionsController.scala
@@ -69,8 +69,8 @@ trait DeductionsController extends FeatureLock {
 
     def routeRequest(totalGain: BigDecimal): Future[Result] = {
       calcConnector.fetchAndGetFormData[PrivateResidenceReliefValueModel](keystoreKeys.prrValue).map {
-        case Some(data) => Ok(views.privateResidenceReliefValue(privateResidenceReliefValueForm(totalGain).fill(data), totalGain, homeLink))
-        case None => Ok(views.privateResidenceReliefValue(privateResidenceReliefValueForm(totalGain), totalGain, homeLink))
+        case Some(data) => Ok(views.privateResidenceReliefValue(privateResidenceReliefValueForm.fill(data), totalGain, homeLink))
+        case None => Ok(views.privateResidenceReliefValue(privateResidenceReliefValueForm, totalGain, homeLink))
       }
     }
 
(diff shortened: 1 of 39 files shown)


app/forms/resident/properties/PrivateResidenceReliefValueForm.scala
```diff
@@ -27,7 +27,7 @@ import uk.gov.hmrc.play.views.helpers.MoneyPounds
 
 object PrivateResidenceReliefValueForm {
 
-  val privateResidenceReliefValueForm = Form(
+  def privateResidenceReliefValueForm(totalGain: BigDecimal): Form[PrivateResidenceReliefValueModel] = Form(
```
DavidWoodward37: Was it intentional to include this here?  Only it isn't used in the validation and I thought we were going to let users enter values higher than their gain?  Then let the backend calculation deal with reducing it down to only 0?

app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -56,15 +57,31 @@ trait DeductionsController extends FeatureLock {
...
 
-  //################# Reliefs Actions ########################
+  //########## Private Residence Relief Actions ##############
+  val privateResidenceReliefValue = FeatureLockForRTT.async { implicit request =>
 
-  def totalGain(answerSummary: YourAnswersSummaryModel, hc: HeaderCarrier): Future[BigDecimal] = calcConnector.calculateRttPropertyGrossGain(answerSummary)(hc)
+    def routeRequest(totalGain: BigDecimal): Future[Result] = {
+      calcConnector.fetchAndGetFormData[PrivateResidenceReliefValueModel](keystoreKeys.prrValue).map {
+        case Some(data) => Ok(views.privateResidenceReliefValue(privateResidenceReliefValueForm(totalGain).fill(data), totalGain, homeLink))
+        case None => Ok(views.privateResidenceReliefValue(privateResidenceReliefValueForm(totalGain), totalGain, homeLink))
```
DavidWoodward37: Are you sure about passing the total gain down into the form?

test/assets/MessageLookup.scala
```diff
@@ -154,6 +154,7 @@ object MessageLookup {
   //Private Residence Relief Value messages
   object privateResidenceReliefValue {
     def title(gain: BigDecimal) = s"How much Private Residence Relief are you claiming on your total gain of £$gain?"
+    val question = "How much Private Residence Relief are you claiming on your total gain of £10,000?"
```
DavidWoodward37: Should this £10,000 not take in a dynamic value? (I think it's alright because we have a gain of 10,000 in all the scala tests, maybe we should change that...)?

PR #458, opened 2016-08-04
Added POST action for PRR value

miranda-hawkes:


app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -80,6 +80,23 @@ trait DeductionsController extends FeatureLock {
...
+    def routeRequest (totalGain: BigDecimal) = {
+      privateResidenceReliefValueForm.bindFromRequest().fold(
+        errors => Future.successful(BadRequest(views.privateResidenceReliefValue(errors, totalGain, homeLink))),
+        success => {
+          calcConnector.saveFormData[PrivateResidenceReliefValueModel](keystoreKeys.prrValue, success)
+          Future.successful(Redirect(routes.DeductionsController.reliefs()))
+        }
+      )
+    }
+    for {
```
jameshforster: Please refactor this into a called function so it only runs when the form has errors.

app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -80,6 +80,23 @@ trait DeductionsController extends FeatureLock {
       route <- routeRequest(totalGain)
     } yield route
   }
+  val submitPrivateResidenceReliefValue = FeatureLockForRTT.async { implicit request =>
+
+    def routeRequest (totalGain: BigDecimal) = {
```
jameshforster: Should be taken out of the function and run so that is calls a different method on a success or error.

jameshforster:
Requires a test in the view for the new added form.

PR #460, opened 2016-08-05
Modified back links to include PRR logic with FeatureLock.

jameshforster:


test/controllers/resident/properties/DeductionsControllerSpec/ReliefsValueActionSpec.scala
```diff
@@ -52,6 +53,7 @@ class ReliefsValueActionSpec extends UnitSpec with WithFakeApplication with Fake
 
     new DeductionsController {
       override val calcConnector: CalculatorConnector = mockCalcConnector
+      val config = mock[AppConfig]
```
DavidWoodward37: My only comment on this whole pr is could we have a consistent method of mocking the app config? Sometimes it's mocked and the set and sometimes it's just mocked and set at the same time (e.g above).

PR #464, opened 2016-08-08
Added PRR to frontend calculation call, result and summaries/reports.

jameshforster:
Merge with CGT-1637


app/connectors/CalculatorConnector.scala
```diff
@@ -214,6 +214,8 @@ trait CalculatorConnector {
   }
 
   def getPropertyDeductionAnswers(implicit hc: HeaderCarrier): Future[ChargeableGainAnswers] = {
+    val privateResidenceReliefModel = fetchAndGetFormData[resident.properties.PrivateResidenceReliefModel](ResidentPropertyKeys.privateResidenceRelief)
```
miranda-hawkes: Being picky, but can you remove the resident.properties by changing the crazy import statement?

PR #490, opened 2016-08-12
Refactored NRCGT Other Reliefs Flat actions to a separate controller.

jameshforster:


DavidWoodward37:
I've just started a re-run of the build.  The test failed due to the 5 second future timeout.

PR #509, opened 2016-08-16
Updated dates to use the no leading zero logic where missing.

jameshforster:


DavidWoodward37:
Could you now delete the datePageFormat val from the Dates object? :)

DavidWoodward37:
Merge conflicts sorry mate

PR #539, opened 2016-08-23
Fixed session timeout pages to use correct calculator routing.

jameshforster:


test/controllers/TimeoutControllerSpec.scala
```diff
@@ -34,7 +34,7 @@ class TimeoutControllerSpec extends UnitSpec with WithFakeApplication {
 
     "when called with no session" should {
 
-      object timeoutTestDataItem extends fakeRequestTo("", TimeoutController.timeout("test"))
+      object timeoutTestDataItem extends fakeRequestTo("", TimeoutController.timeout("test", "test2"))
```
DavidWoodward37: Could you add a little test for the home link being assigned correctly?

PR #551, opened 2016-09-02
Changes to Improvements page as in prototype

morgan-fernandes:


app/views/calculation/resident/properties/gain/improvements.scala.html
```diff
@@ -1,14 +1,16 @@
...
+@import views.html.helpers.resident._
 @import views.html.calculation.resident.properties._
 
 @(improvementsForm : Form[ImprovementsModel])(implicit request: Request[_])
 
-@note = {
-    <div class="panel panel-border-wide">
-        <p>@Messages("calc.resident.improvements.note")</p>
-    </div>
+@hiddenHelpTextContent = {
```
macgrewal: The 2 messages in hiddenHelpTextContent should both be in `p` tags like
```
@hiddenHelpTextContent = {
    <p>@Messages("calc.resident.improvements.additionalContent.one")</p>
    <p>@Messages("calc.resident.improvements.additionalContent.two")</p>
}
```

PR #557, opened 2016-09-06
Added new Lettings Relief page

miranda-hawkes:
Merge with acceptance tests
Also stubbed a few pages to minimise rework when other new pages are added in


app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -70,6 +71,41 @@ trait DeductionsController extends FeatureLock {
...
+    def errorAction(form: Form[LettingsReliefModel]) = {
+      Future.successful(BadRequest(views.lettingsRelief(form, homeLink, Some(lettingsReliefBackUrl))))
+    }
+
+    def successAction(model: LettingsReliefModel) = {
+      calcConnector.saveFormData[LettingsReliefModel](keystoreKeys.lettingsRelief, model)
+      model match {
+        case LettingsReliefModel(true) => Future.successful(Redirect(routes.DeductionsController.lettingsReliefValue()))
+        case _ => Future.successful(Redirect(routes.DeductionsController.otherProperties()))
+      }
```
DavidWoodward37: You could re-write this to an async function.  (so have the route selection and the save data in separate threads).

app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -70,6 +71,41 @@ trait DeductionsController extends FeatureLock {
...
+
+  //############## Lettings Relief Actions ##################
+
+  private val lettingsReliefBackUrl = routes.DeductionsController.privateResidenceReliefValue().url
+
+  val lettingsRelief = FeatureLockForRTT.async { implicit request =>
+    calcConnector.fetchAndGetFormData[LettingsReliefModel](keystoreKeys.lettingsRelief).map {
+      case Some(data) => Ok(views.lettingsRelief(lettingsReliefForm.fill(data), homeLink, Some(lettingsReliefBackUrl)))
+      case None => Ok(views.lettingsRelief(lettingsReliefForm, homeLink, Some(lettingsReliefBackUrl)))
+    }
```
DavidWoodward37: Out of curiosity if the back link was static why did we not put it in the view?

PR #565, opened 2016-09-06
Properties introduction page added and home link navigation fixed

macgrewal:


macgrewal:
Back link needs adding to the disposal date page

DavidWoodward37:
Holding off merging it then ^_^

conf/messages
```diff
@@ -298,6 +298,13 @@ calc.summary.warning.taxYear = Your total might be less accurate because you did
 ## Resident messages ##
 ###############################################################################################
 
+## Introduction messages
+calc.resident.properties.introduction.title = Work out how much Capital Gains Tax you owe
+calc.resident.properties.introduction.subheading = Do you need to use this calculator?
+calc.resident.properties.introduction.paragraph = You probably don''t need to pay Capital Gains Tax if the property you''ve sold is your own home. You''ll be entitled to a tax relief called Private Residence Relief.
+calc.resident.properties.introduction.entitledLinkText = Find out if you''re entitled to Private Residence Relief (opens in a new window).
```
DavidWoodward37: Would it be worth us doing a whole re-factor across the project?  We currently always split the text from the sentence and the (opens in a new window text).  Mostly because it was all in place and then we simply removed the visually hidden classes but not the enclosing span tags around (opens in a new window)?

test/views/resident/properties/IntroductionViewSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+    lazy val view = views.introduction()(fakeRequest)
+    lazy val doc = Jsoup.parse(view.body)
+
+    def scopedSelector(selector: String): String = s"article.content__body $selector"
+
+    val headingSelector = scopedSelector("h1")
+    val subheadingSelector = scopedSelector("h2")
+    val helpTextSelector = scopedSelector("p:nth-of-type(1)")
+    val entitlementLinkSelector = scopedSelector("a:nth-of-type(1)")
+    val continueLinkSelector = scopedSelector("a:nth-of-type(2)")
```
DavidWoodward37: Out of curiosity was there any particular reason for changing to this method of selecting rather than have say: 
lazy val article = doc.select("article.content__body") 
and then subsequent 
article.select("h2")... ?

DavidWoodward37:
The routing on the outside-tax-years page still has the home link as heading back to the disposal-date page.  Should it not go back to the introduction page now?

PR #570, opened 2016-09-07
CGT-1841: Added the Letting Relief Value to the Summaries and Reports

MJCallahanPage:


app/views/calculation/resident/properties/report/deductionsSummaryReport.scala.html
```diff
@@ -139,13 +139,16 @@ <h1 class="heading-xlarge">
...
+        @{if(deductionAnswers.propertyLivedInModel.get.livedInProperty)
         summaryOptionRowHelper("privateResidenceRelief", Messages("calc.resident.properties.privateResidenceRelief.title"),
         deductionAnswers.privateResidenceReliefModel.get.isClaiming, None)}
-        @{if (deductionAnswers.propertyLivedInModel.get.livedInProperty && deductionAnswers.privateResidenceReliefModel.get.isClaiming)
-        summaryOptionRowHelper("lettingsRelief", Messages("calc.resident.lettingsRelief.title"),
-        deductionAnswers.lettingsReliefModel.get.isClaiming, None)}
-
+        @if(deductionAnswers.propertyLivedInModel.get.livedInProperty && deductionAnswers.privateResidenceReliefModel.get.isClaiming){
+            @summaryOptionRowHelper("lettingsRelief", Messages("calc.resident.lettingsRelief.title"),deductionAnswers.lettingsReliefModel.get.isClaiming, None)
+            @if(deductionAnswers.lettingsReliefModel.get.isClaiming){
```
DavidWoodward37: I'm not sure about these nested if's.  Having content that is conditional on answers to prior questions is alright but (whilst it's doing that.) this is almost structured to say this content doesn't appear without the other content.  Maybe one to discuss for the team?

DavidWoodward37:
MERGE CONFLICTS!!! :D

app/views/calculation/resident/properties/report/deductionsSummaryReport.scala.html
MJCallahanPage: Yeah I don't like the logic we have in the view - I think it needs pulling out into a constructor.

However, the structure is correct, the content will never appear without the other content as the only way you could get to Lettings Relief is if you've gone through Live in Property and PRR Entitlement. So in a way, it's correct.

MJCallahanPage:
@DavidWoodward37 - Conflicts resolved. As we discussed I've also taken some steps to update the Summary code structure. It's still messy and we need to consider raising a Technical Debt task to investigate better ways of generating the Summary pages with less logic in the view itself :smile:

PR #577, opened 2016-09-13
Added tests for all GET requests in nr.routes

euptonqa:


test/routes/nrcgt/RoutesSpec.scala
```diff
@@ -0,0 +1,197 @@
...
+      path shouldEqual "/calculate-your-capital-gains/non-resident/other-reliefs-time-apportioned"
+    }
+  }
+
+  "The URL for the other reliefs rebased Action" should {
+    "be equal to /calculate-your-capital-gains/non-resident/other-reliefs-rebased" in {
+      val path = controllers.nonresident.routes.OtherReliefsRebasedController.otherReliefsRebased().url.toString
+      path shouldEqual "/calculate-your-capital-gains/non-resident/other-reliefs-rebased"
+    }
+  }
```
DavidWoodward37: Sorry but could you also add tests for all the submit actions as well? The URL's they POST to are identical as the GET's but both should be tested.

PR #579, opened 2016-09-14
Added Sell For Less Page

morgan-fernandes:


test/controllers/resident/properties/GainControllerSpec/SellForLessActionSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+
+      lazy val target = setupTarget(Some(SellForLessModel(true)))
+      lazy val result = target.sellForLess(fakeRequestWithSession)
+
+      "return a status of 200" in {
+        status(result) shouldBe 200
+      }
+
+      s"return some html with title of ${messages.title}" in {
+        contentType(result) shouldBe Some("text/html")
```
DavidWoodward37: If you're going to keep this test for the HTML it's worth spliting this into two separate assertions, one for the html and one for the title.

test/controllers/resident/properties/GainControllerSpec/SellForLessActionSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+      lazy val target = setupTarget(None)
+      lazy val request = fakeRequestToPOSTWithSession(("sellForLess", "Yes"))
+      lazy val result = target.submitSellForLess(request)
+
+      "return a status of 303" in {
+        status(result) shouldBe 303
+      }
+
+      "redirect to the worth when sold page" in {
+        redirectLocation(result) shouldBe Some("/calculate-your-capital-gains/resident/properties/worth-when-sold")
```
DavidWoodward37: I'll update this route in my pull request.

test/views/resident/properties/gain/SellForLessViewSpec.scala
```diff
@@ -0,0 +1,233 @@
...
+      }
+
+      "has a back link to 'back'" in {
+        backLink.attr("href") shouldBe "back-link"
+      }
+    }
+
+    "render a form tag with a submit action" in {
+      doc.select("form").attr("action") shouldEqual "/calculate-your-capital-gains/resident/properties/sell-for-less"
+    }
```
DavidWoodward37: It might be worth adding another test here as this will pass even if the method uses the GET.

test/views/resident/properties/gain/SellForLessViewSpec.scala
```diff
@@ -0,0 +1,233 @@
...
+
+    "have a legend for the radio inputs" which {
+
+      lazy val legend = doc.select("legend")
+
+      s"contain the text ${messages.title}" in {
+        legend.text should include(s"${messages.title}")
+      }
```
DavidWoodward37: Maybe check that the legend is also visually hidden?

PR #581, opened 2016-09-15
Added Worth When Gave Away page

morgan-fernandes:


conf/messages
```diff
@@ -331,6 +331,10 @@ calc.resident.properties.sellForLess.noSelectError = Tell us if you have ever so
 calc.resident.sellOrGiveAway.title = Did you sell the property or give it away?
 calc.resident.sellOrGiveAway.errors.mandatory = Tell us if you sold the property or gave it away
 
+##Worth When Gave Away messages
+calc.resident.properties.worthWhenGaveAway.title = What was the property worth when you gave it away?
+calc.resident.properties.worthWhenGaveAway.question = What was the property worth when you gave it away?
```
DavidWoodward37: We had a discussion with mac about removing all these double entries for now, originally it because on some pages the title and the question was different, but it's now pointless to have them both, just keep the title maybe?

test/assets/MessageLookup.scala
```diff
@@ -387,4 +387,9 @@ object MessageLookup {
     val title = "Did you sell the property or give it away?"
     val errorMandatory = "Tell us if you sold the property or gave it away"
   }
+
+  object propertiesWorthWhenGaveAway {
+    val title = "What was the property worth when you gave it away?"
+    val question = "What was the property worth when you gave it away?"
```
DavidWoodward37: Again removing the duplicated text would be good for now.

test/assets/MessageLookup.scala
```diff
@@ -387,4 +387,9 @@ object MessageLookup {
...
+  }
```
DavidWoodward37: Could you move this entire object into the new Resident Properties nested objects at the top of the message lookup file.  It will mean slightly less refactoring later (and the structures already there).

test/controllers/resident/properties/GainControllerSpec/WorthWhenGaveAwayActionSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  "Calling .submitWorthWhenGaveAway from the GainController" should {
+    lazy val target = setupTarget(None)
+    lazy val request = fakeRequestToPOSTWithSession(("amount", "100"))
+    lazy val result = target.submitWorthWhenGaveAway(request)
+
+    "re-direct to the disposal Costs page when supplied with a valid form" in {
+      status(result) shouldEqual 303
+      redirectLocation(result) shouldBe Some("/calculate-your-capital-gains/resident/properties/disposal-costs")
```
DavidWoodward37: Can you split these asserts into two separate tests?

test/controllers/resident/properties/GainControllerSpec/WorthWhenGaveAwayActionSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  "Calling .submitWorthWhenGaveAway from the GainController" should {
+    lazy val target = setupTarget(None)
+    lazy val request = fakeRequestToPOSTWithSession(("amount", ""))
+    lazy val result = target.submitWorthWhenGaveAway(request)
+
+    "render the worth when gave away page when supplied with an invalid form" in {
+      status(result) shouldEqual 400
+      Jsoup.parse(bodyOf(result)).title() shouldEqual MessageLookup.propertiesWorthWhenGaveAway.title
```
DavidWoodward37: And here too

test/views/resident/properties/gain/WorthWhenGaveAwayViewSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    s"have a back link to the Who did you give it to Page with text ${MessageLookup.calcBaseBack}" in {
+      doc.select("#back-link").attr("href") shouldEqual "/calculate-your-capital-gains/resident/properties/who-did-you-give-it-to"
+    }
+
+    s"have the question of the page ${messages.question}" in {
+      doc.select("h1").text shouldEqual messages.question
+    }
+
```
DavidWoodward37: I think there are tests missing here around the actual input on the page, it's hidden label and the label having the same text as the title/question (which ever you decide to keep).  Ideally this would have been done in a helper but we'd still need the hidden text tested as that is an option for the helper not a built in feature.

PR #587, opened 2016-09-16
Added Worth On Page

morgan-fernandes:


DavidWoodward37:
Sorry morgan, probably import merge conflicts in the Gain Controller.

DavidWoodward37:
The build failed as well but it was due to a future timing out.  I've restarted and it should pass now.

app/views/calculation/resident/properties/gain/worthOn.scala.html
```diff
@@ -0,0 +1,27 @@
+@import uk.gov.hmrc.play.views.html.helpers.form
+@import models.resident.properties._
+@import views.html.helpers._
+@import views.html.calculation.resident._
+
+@(worthOnForm: Form[WorthOnModel], navBackLink: Option[String], navHomeLink: String, postAction: Call)(implicit request: Request[_])
+
+
+@resident_main_template(
```
macgrewal: use `resident_properties_main_template` instead of `resident_main_template` this will reduce the inputs to the template

app/views/calculation/resident/properties/gain/worthOn.scala.html
```diff
@@ -0,0 +1,27 @@
...
+backLink = navBackLink,
+homeLink = navHomeLink,
+navTitle = ""
+){
+
+@errorSummary(worthOnForm, "amount")
+
+<h1 class="heading-large">@Messages("calc.resident.properties.worthOn.title")</h1>
+
+@form(action = postAction){
```
macgrewal: specify the post action in the view, this page is not shared with shares so the post action will always be the same

conf/messages
```diff
@@ -368,6 +368,10 @@ calc.resident.howBecameOwner.bought = Bought it
 calc.resident.howBecameOwner.gifted = Got it as a gift
 calc.resident.howBecameOwner.inherited= Inherited it
 
+
+##Worth On messages
+calc.resident.properties.worthOn.title = What was the property worth on 31 March 1982?
```
macgrewal: would make more sense to call the key `calc.resident.properties.worthOn.question` as it is the question. It's used as the title but it is in fact a question.

conf/resident_properties.routes
```diff
@@ -43,13 +43,14 @@ POST    /disposal-costs                   controllers.resident.properties.GainCo
...
 
-#Property Worth on 31/03/1982 Routes
-GET     /property-worth-in-may            controllers.resident.properties.GainController.propertyWorthInMayEightyTwo
-
 #How Became Owner Routes
 GET     /how-became-owner                 controllers.resident.properties.GainController.howBecameOwner
 POST    /how-became-owner                 controllers.resident.properties.GainController.submitHowBecameOwner
 
+#Worth On Routes
+GET     /worth-on                         controllers.resident.properties.GainController.worthOn
```
macgrewal: A more meaningfull URL would be /market-value-on-31-march-1982

PR #600, opened 2016-09-21
Added bought-for-less-than-worth question to resident property summaries and reports.

jameshforster:


DavidWoodward37:
Time for some MERGE CONFLICTS :dancer:

PR #604, opened 2016-09-22
Added worth-when-bought question to resident property summaries and reports.

jameshforster:


app/connectors/CalculatorConnector.scala
```diff
@@ -211,6 +211,10 @@ trait CalculatorConnector {
       case Some(data) => Some(data.boughtForLessThanWorth)
       case None => None
     }
+    val worthWhenBought = fetchAndGetFormData[properties.WorthWhenBoughtModel](ResidentPropertyKeys.worthWhenBought).map {
```
DavidWoodward37: Can you update the name of this to worthWhenBoughtForLess? (And in all the places it's used) I think it makes more sense for this parameter.

app/models/resident/properties/YourAnswersSummaryModel.scala
```diff
@@ -33,7 +33,8 @@ case class YourAnswersSummaryModel
   sellForLess: Option[Boolean],
   ownerBeforeAprilNineteenEightyTwo: Boolean,
   howBecameOwner: Option[String],
-  boughtForLessThanWorth: Option[Boolean]
+  boughtForLessThanWorth: Option[Boolean],
+  worthWhenBought: Option[BigDecimal]
```
DavidWoodward37: Name update please!

DavidWoodward37:
And some merge conflicts!

PR #612, opened 2016-09-23
CGT-1952: Added the validation for the Letting Relief

MJCallahanPage:
The validation is quite horrible to ensure that only the one error message is displayed based on whichever is the lowest value. Someone could probably refactor this to be more clever.

### DIFF
--- a/app/controllers/resident/properties/DeductionsController.scala
+++ b/app/controllers/resident/properties/DeductionsController.scala
@@ -215,35 +215,38 @@ trait DeductionsController extends FeatureLock {
   //################# Lettings Relief Value Input Actions ########################
   val lettingsReliefValue = FeatureLockForRTT.async { implicit request =>
 
-    def routeRequest(totalGain: BigDecimal): Future[Result] = {
+    def routeRequest(totalGain: BigDecimal, prrValue: BigDecimal): Future[Result] = {
       calcConnector.fetchAndGetFormData[LettingsReliefValueModel](keystoreKeys.lettingsReliefValue).map {
-        case Some(data) => Ok(views.lettingsReliefValue(lettingsReliefValueForm.fill(data), homeLink, totalGain))
-        case None => Ok(views.lettingsReliefValue(lettingsReliefValueForm, homeLink, totalGain))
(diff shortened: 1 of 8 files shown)


conf/messages
```diff
@@ -460,7 +460,9 @@ calc.resident.properties.privateResidenceReliefValue.gainExceededError = Enter a
 calc.resident.lettingsReliefValue.title = How much Letting Relief are you entitled to?
 calc.resident.lettingsReliefValue.question = How much Letting Relief are you entitled to?
 calc.resident.lettingsReliefValue.additionalContent = We''ve calculated that you''ve made a gain of £{0} on your property. You''ll need this figure to calculate your Letting Relief.
-
+calc.resident.lettingsReliefValue.error.moreThanCappedAmount = The Letting Relief you''ve entered is more than the maximum amount of £{0}
+calc.resident.lettingsReliefValue.error.moreThanPrr = The Letting Relief amount you''ve entered is more than your Private Residence Relief
+calc.resident.lettingsReliefValue.error.moreThanRemainingGain = The Letting Relief you''ve entered is more than your remaining gain of £{0}
```
DavidWoodward37: Were these the messages you validated with Matt? (They break from our current style of messages which are more instructional rather than just a statement of what's wrong.)

app/controllers/resident/properties/DeductionsController.scala
```diff
@@ -215,35 +215,38 @@ trait DeductionsController extends FeatureLock {
...
+        success => {
+          calcConnector.saveFormData[LettingsReliefValueModel](keystoreKeys.lettingsReliefValue, success)
+          Future.successful(Redirect(routes.DeductionsController.otherProperties()))
+        })
     }
 
-    lettingsReliefValueForm.bindFromRequest().fold(errorAction, successAction)
+    for {
+      answerSummary <- answerSummary(hc)
+      totalGain <- totalGain(answerSummary, hc)
```
DavidWoodward37: There is currently a 'bug' here that I think will be fixed when the calculation call is updated (because when the totalGain method is called at the moment regardless of whether it was inherited or gifted or bought the bought value is always used.)  This should be fixed by updating the calculation call though.

PR #642, opened 2016-10-12
Added the view and spec for the mandatory rebased value page.

DavidWoodward37:


conf/messages
```diff
@@ -151,6 +151,7 @@ calc.acquisitionValue.bullet.five = became the owner before 1 April 1982
 calc.rebasedValue.question = Do you know how much the property was worth on 5 April 2015?
 calc.rebasedValue.questionTwo = What was the value of the property?
 calc.rebasedValue.helpText = Only tell us if you owned the property on that date
+calc.nonResident.rebasedValue.inputHelpText = You can use a valuation from a surveyor or a property website.
```
morgan-fernandes: The question does not match the prototype.

test/assets/MessageLookup.scala
```diff
@@ -41,6 +41,19 @@ object MessageLookup {
...
   // to the right place in this object.
 
+  object NonResident {
+
+    object Common {
+      val pageHeading = "Calculate your non-resident Capital Gains Tax"
+    }
+
+    object RebasedValue {
+      val question = "What was the value of the property?"
```
morgan-fernandes: The text would need to be updated for the question.

PR #651, opened 2016-10-14
Updated the deduction summary and report to utilise predicates found in Chargeable Gain model

euptonqa:


jameshforster:
Raised against master, needs to be raised against CGT-1874

PR #654, opened 2016-10-14
Updated the deduction summary and report to utilise predicates found in Deduction Answers model for shares

euptonqa:


app/views/calculation/resident/shares/report/deductionsSummaryReport.scala.html
```diff
@@ -146,15 +146,15 @@ <h1 class="heading-xlarge">
         @{if (deductionAnswers.otherPropertiesModel.get.hasOtherProperties)
```
jameshforster: Has not updated any of the earlier logic to use the boolean predicates (from total gain), check the big nested if and remove it for the simpler logic.

app/views/calculation/resident/shares/summary/deductionsSummary.scala.html
```diff
@@ -168,12 +168,12 @@ <h1 class="heading-xlarge">
     @summaryOptionRowHelper("otherDisposals",Messages("calc.resident.otherProperties.title", taxYear.taxYearSupplied),deductionAnswers.otherPropertiesModel.get.hasOtherProperties,Some(deductionRoutes.otherDisposals().toString))
```
jameshforster: Has not updated any of the earlier logic to use the boolean predicates (from total gain), check the big nested if and remove it for the simpler logic.

PR #662, opened 2016-10-18
Updated Personal details section tests to use MessageLookup object

TomDudley92:


conf/messages
```diff
@@ -105,7 +105,7 @@ calc.personalAllowance.linkOne = Income Tax rates and Personal Allowances
 calc.personalAllowance.linkTwo = Tax on your UK income if you live abroad
 calc.personalAllowance.errorNegative = Enter a positive number for your Personal Allowance
 calc.personalAllowance.errorDecimalPlaces = Enter a whole number for your Personal Allowance
-calc.personalAllowance.errorMaxLimit = Enter a Personal Allowance that’s £
+calc.personalAllowance.errorMaxLimit = Enter a Personal Allowance that''s £
```
euptonqa: Will this not cause issues? I don't think the ' needs to be escaped and in the previous messages there's only one ', not two.

PR #663, opened 2016-10-18
Updated messages in NRCGT Improvements View Spec

euptonqa:


DavidWoodward37:
Heya, sorry the build failed SummarySpec.scala:352, you'll need to change the assert in the test from shouldEqual/shouldBe to should include(....)

PR #674, opened 2016-10-21
Finished the PersonalDetailsConstructor

TomDudley92:


app/constructors/nonresident/PersonalDetailsConstructor.scala
```diff
@@ -0,0 +1,131 @@
...
+      Messages("calc.otherProperties.question"),
+      Some(controllers.nonresident.routes.OtherPropertiesController.otherProperties().url)
+    ))
+  }
+
+  def getOtherPropertiesAmountAnswer(summaryModel: SummaryModel): Option[QuestionAnswerModel[BigDecimal]] =
+    (summaryModel.otherPropertiesModel.otherProperties,
+      summaryModel.otherPropertiesModel.otherPropertiesAmt,
+      summaryModel.customerTypeModel.customerType) match {
+    case ("Yes", Some(otherPropertiesAmount), CustomerTypeKeys.individual) => Some(QuestionAnswerModel(
```
jameshforster: This should not require a CustomerType, this is possible for all customers.

app/constructors/nonresident/PersonalDetailsConstructor.scala
```diff
@@ -0,0 +1,131 @@
...
+      KeystoreKeys.otherProperties,
+      summaryModel.otherPropertiesModel.otherProperties,
+      Messages("calc.otherProperties.question"),
+      Some(controllers.nonresident.routes.OtherPropertiesController.otherProperties().url)
+    ))
+  }
+
+  def getOtherPropertiesAmountAnswer(summaryModel: SummaryModel): Option[QuestionAnswerModel[BigDecimal]] =
+    (summaryModel.otherPropertiesModel.otherProperties,
+      summaryModel.otherPropertiesModel.otherPropertiesAmt,
```
jameshforster: Would it not make more sense to match with a model instead of both components separately?

test/constructors/nonresident/PersonalDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,382 @@
...
+        PersonalDetailsConstructor.getPersonalDetailsSection(summaryWithAllOptionValuesModel)
+          .count(_.equals(PersonalDetailsConstructor.getOtherPropertiesAnswer(summaryWithAllOptionValuesModel).get)) shouldBe 1
+      }
+
+      ".getPurchaseDetailsItem will return a Sequence[QuestionAnswerModel[Any]] will contain a PersonalDetails.getOtherPropertiesAmountAnswer item" in {
+        PersonalDetailsConstructor.getPersonalDetailsSection(summaryWithAllOptionValuesModel)
+          .count(_.equals(PersonalDetailsConstructor.getOtherPropertiesAmountAnswer(summaryWithAllOptionValuesModel).get)) shouldBe 1
+      }
+
+      ".getPurchaseDetailsItem will return a Sequence[QuestionAnswerModel[Any]] will not contain anything other than " +
```
jameshforster: This test is not actually necessary as you have already checked the size and the components of the result

test/constructors/nonresident/PersonalDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,382 @@
...
+    OtherReliefsModel(None, None),
+    None
+  )
+
+  "Calling PersonalDetailsConstructor" when {
+
+    "using the summaryWithAllOptionsValuesModel" should {
+
+      ".getPurchaseDetailsItem will return a Sequence[QuestionAnswerModel[Any]] with size 5" in{
+        PersonalDetailsConstructor.getPersonalDetailsSection(summaryWithAllOptionValuesModel).size shouldBe 5
```
euptonqa: Just in case I forget to fix this - the test is referring to non-existent getPurchaseDetailsItem function but calling getPersonalDetailsSection function. Same applies to related tests.

PR #681, opened 2016-10-26
PurchaseDetails Constructor

euptonqa:


test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,280 @@
...
+
+    "using the summaryWithAllOptionsValuesModel" should {
+      ".getPurchaseDetailsSection will return a Sequence[QuestionAnswerModel[Any]] with size 5" in {
+        PurchaseDetailsConstructor
+          .getPurchaseDetailsSection(summaryWithAllOptionValuesModel).size shouldBe 5
+      }
+
+      ".getPurchaseDetailsSection will return a Sequence[QuestionAnswerModel[Any]] that will contain an acquisitionDateData item" in {
+        PurchaseDetailsConstructor
+          .getPurchaseDetailsSection(summaryWithAllOptionValuesModel)
```
jameshforster: Can be replaced with .contains instead of .count(_.equals) and that returns a boolean

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,280 @@
...
+    DisposalCostsModel(5000.0),
+    AllowableLossesModel("No", None),
+    CalculationElectionModel("flat"),
+    OtherReliefsModel(None, None),
+    OtherReliefsModel(None, None),
+    OtherReliefsModel(None, None),
+    None
+  )
+
+  "Calling the PurchaseDetailsConstructor" when {
```
jameshforster: The constructor is never called, sort into calls to the functions.
Please look at the DeductionDetailsControllerSpec as an example of how to layout your tests

app/constructors/nonresident/PurchaseDetailsConstructor.scala
```diff
@@ -0,0 +1,98 @@
...
+    Some(QuestionAnswerModel(
+      KeystoreKeys.acquisitionCosts,
+      summaryModel.acquisitionCostsModel.acquisitionCostsAmt,
+      Messages("calc.acquisitionCosts.question"),
+      Some(controllers.nonresident.AcquisitionCostsController.acquisitionCosts.toString())
+    ))
+  }
+
+  def getRebasedValueAnswer(summaryModel: SummaryModel): Option[QuestionAnswerModel[Option[BigDecimal]]] =
+    (summaryModel.rebasedValueModel) match {
```
jameshforster: Needs to check whether calculation type is rebased (same for Costs below) otherwise return a None

app/constructors/nonresident/PurchaseDetailsConstructor.scala
```diff
@@ -0,0 +1,98 @@
...
+    Some(QuestionAnswerModel(
+      KeystoreKeys.acquisitionDate,
+      Dates.constructDate(summaryModel.acquisitionDateModel.day.get,
+        summaryModel.acquisitionDateModel.month.get, summaryModel.acquisitionDateModel.year.get),
+      Messages("calc.acquisitionDate.questionTwo"),
+      Some(controllers.nonresident.AcquisitionDateController.acquisitionDate().toString())
+    ))
+  }
+
+  def getAcquisitionValueAnswer(summaryModel: SummaryModel): Option[QuestionAnswerModel[BigDecimal]] =  {
```
jameshforster: Needs to check that calculation type is NOT rebased first same for costs below otherwise return a None

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,280 @@
...
+import java.text.DateFormat
+import java.time.{LocalDate, LocalTime}
+
+import common.nonresident.CustomerTypeKeys
+import models.nonresident._
+import common.KeystoreKeys
+import uk.gov.hmrc.play.test.{UnitSpec, WithFakeApplication}
+import constructors.nonresident.PurchaseDetailsConstructor
+
+/**
```
jameshforster: Remove this comment please

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,280 @@
...
+  * Created by emma on 24/10/16.
+  */
+object PurchaseDetailsConstructorSpec extends UnitSpec with WithFakeApplication {
```
jameshforster: This needs to be a class, will not run as an object

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,280 @@
...
+          .getPurchaseDetailsSection(summaryWithAllOptionValuesModel)
+          .count(_.equals(PurchaseDetailsConstructor
+            .getRebasedCostsAnswer(summaryWithAllOptionValuesModel))) shouldBe 1
+      }
+
+      ".getAcquisitionDateAnswer will return a date of 4/9/16" in {
+        val obtainedDate: LocalDate = PurchaseDetailsConstructor.getAcquisitionDateAnswer(summaryWithAllOptionValuesModel).get.data
+        val stringConversion = obtainedDate.getDayOfMonth + "/" + obtainedDate.getMonth +
+          "/" + obtainedDate.getYear
+        PurchaseDetailsConstructor.getAcquisitionDateAnswer(summaryWithAllOptionValuesModel) shouldBe stringConversion
```
jameshforster: None of these tests are checking for a QuestionAnswerModel so all of them are failing (the values are not the outputs)

app/constructors/nonresident/PurchaseDetailsConstructor.scala
```diff
@@ -44,36 +44,42 @@ object PurchaseDetailsConstructor {
...
   }
 
   def getAcquisitionValueAnswer(summaryModel: SummaryModel): Option[QuestionAnswerModel[BigDecimal]] =  {
-    Some(QuestionAnswerModel(
-      KeystoreKeys.acquisitionValue,
-      summaryModel.acquisitionValueModel.acquisitionValueAmt,
-      Messages("calc.acquisitionValue.question"),
-      Some(controllers.nonresident.AcquisitionValueController.acquisitionValue().toString())
-    ))
+    if (summaryModel.calculationElectionModel.calculationType.equals(CalculationType.flat))
```
jameshforster: This is not only true for a flat calculation, it needs to be true for a time apportioned one too (e.g. when !rebased)

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -81,200 +79,186 @@ object PurchaseDetailsConstructorSpec extends UnitSpec with WithFakeApplication
...
-          .getRebasedCostsAnswer((summaryWithAllOptionValuesModel)).get.data
-        obtainedValue shouldBe 4000.0
-      }
-
-      ".getAcquisitionDateAnswer will return a value of 4/9/16" in {
-        val obtainedDate = PurchaseDetailsConstructor
-          .getAcquisitionDateAnswer(summaryWithAllOptionValuesModel).get.data
-        val stringConversion = obtainedDate.getDayOfMonth + "/" + obtainedDate.getMonth +
+        val obtainedDate: LocalDate = PurchaseDetailsConstructor.getAcquisitionDateAnswer(summaryFlatOptionValuesModel).get.data
+        val stringConversion = obtainedDate.getDayOfMonth() + "/" + obtainedDate.getMonthValue() +
```
jameshforster: Can we check this against a simulated date item instead of converting to a string manually?
e.g. obtainedDate shouldBe LocalDate.parse("2016-09-04")

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -81,200 +79,186 @@ object PurchaseDetailsConstructorSpec extends UnitSpec with WithFakeApplication
...
-            .getRebasedValueAnswer(summaryWithAllOptionValuesModel))) shouldBe 1
-      }
 
-      ".getPurchaseDetailsSection will return a Sequence[QuestionAnswersModel[Any]] that will contain a rebasedCostsData item" in {
-        PurchaseDetailsConstructor
-          .getPurchaseDetailsSection(summaryWithAllOptionValuesModel)
-          .count(_.equals(PurchaseDetailsConstructor
-            .getRebasedCostsAnswer(summaryWithAllOptionValuesModel))) shouldBe 1
-      }
+      /** ***/
```
jameshforster: None of the get$ROWNAME tests are checking the question message, please add these

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -81,200 +79,186 @@ object PurchaseDetailsConstructorSpec extends UnitSpec with WithFakeApplication
   "Calling the PurchaseDetailsConstructor" when {
 
     "using the summaryWithAllOptionsValuesModel" should {
```
jameshforster: As this method is responsible for calling each row method, you need to check that in the scenario where a rebased value is given that the rebased rows appear.

app/constructors/nonresident/PurchaseDetailsConstructor.scala
```diff
@@ -0,0 +1,104 @@
...
+        KeystoreKeys.acquisitionDate,
+        Dates.constructDate(summaryModel.acquisitionDateModel.day.get,
+          summaryModel.acquisitionDateModel.month.get, summaryModel.acquisitionDateModel.year.get),
+        Messages("calc.acquisitionDate.questionTwo"),
+        Some(controllers.nonresident.AcquisitionDateController.acquisitionDate().toString())
+      ))
+  }
+
+  def getAcquisitionValueAnswer(summaryModel: SummaryModel): Option[QuestionAnswerModel[BigDecimal]] =  {
+    if (summaryModel.calculationElectionModel.calculationType.equals(CalculationType.flat))
```
jameshforster: This is not only true for a flat calculation, it needs to be true for a time apportioned one too (e.g. when !rebased)

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+        val obtainedValue = PurchaseDetailsConstructor
+          .getRebasedCostsAnswer((summaryFlatOptionValuesModel))
+        obtainedValue shouldBe None
+      }
+    }
+
+    "using the summaryRebasedOptionsModel" should {
+
+      "using the summaryWithAllOptionsValuesModel" should {
+        ".getPurchaseDetailsSection with a rebased election type model" +
```
jameshforster: The section needs to check that it is calling the rebased values correctly in this scenario (same as when you checked the acquisition value/costs before)

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+    CalculationElectionModel("rebased"),
+    OtherReliefsModel(None, None),
+    OtherReliefsModel(None, None),
+    OtherReliefsModel(None, None),
+    None
+  )
+
+  "Calling the PurchaseDetailsConstructor" when {
+
+    "using the summaryWithAllOptionsValuesModel" should {
```
jameshforster: There are no tests for the messages anywhere, this is important as it is passed in by the constructor. Please add them to each getRow method

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+      }
+
+
+      /** ***/
+
+      ".getAcquisitionDateAnswer will return a date of 4/9/16" in {
+        val obtainedDate: LocalDate = PurchaseDetailsConstructor.getAcquisitionDateAnswer(summaryFlatOptionValuesModel).get.data
+        val stringConversion = obtainedDate.getDayOfMonth() + "/" + obtainedDate.getMonthValue() +
+          "/" + obtainedDate.getYear
+        stringConversion shouldBe "4/9/2016"
```
jameshforster: Can we check this against a simulated date item instead of converting to a string manually?
e.g. obtainedDate shouldBe LocalDate.parse("2016-09-04")

test/constructors/nonresident/PurchaseDetailsConstructorSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+    OtherReliefsModel(None, None),
+    None
+  )
+
+  "Calling the PurchaseDetailsConstructor" when {
+
+    "using the summaryWithAllOptionsValuesModel" should {
+      ".getPurchaseDetailsSection with a flat election type model" +
+        " will return a Sequence[QuestionAnswerModel[Any]] with size 3" in {
+        PurchaseDetailsConstructor
```
jameshforster: Methods are called multiple times within the same scope (e.g. PurchaseDetailsConstructor.getPurchaseDetailsSection(summaryFlatOptionValuesModel))

Would it not make more sense to move these into a single val rather than repeating code?

PR #691, opened 2016-10-31
Refactored tests into view, form and action

TomDudley92:


app/forms/nonresident/DisposalCostsForm.scala
```diff
@@ -25,11 +26,14 @@ import play.api.i18n.Messages
 import uk.gov.hmrc.play.views.helpers.MoneyPounds
 
 object DisposalCostsForm {
-  val disposalCostsForm = Form(
+  def disposalCostsForm(disposalCosts: BigDecimal = BigDecimal(0)): Form[DisposalCostsModel] = Form(
```
jameshforster: Revert to a val without variables please, they are not required for this form.

test/controllers/CalculationControllerTests/DisposalCostsActionSpec.scala
```diff
@@ -232,8 +186,8 @@ class DisposalCostsSpec extends UnitSpec with WithFakeApplication with MockitoSu
         status(result) shouldBe 400
```
jameshforster: Only one 400 response needs to be tested (as we aren't testing each failure case, the form does that), please remove all extra ones.

test/controllers/CalculationControllerTests/DisposalCostsActionSpec.scala
```diff
@@ -38,7 +38,7 @@ import models.nonresident.{AcquisitionDateModel, DisposalCostsModel, RebasedValu
 import play.api.mvc.Result
 import uk.gov.hmrc.play.views.helpers.MoneyPounds
 
-class DisposalCostsSpec extends UnitSpec with WithFakeApplication with MockitoSugar {
+class DisposalCostsActionSpec extends UnitSpec with WithFakeApplication with MockitoSugar {
 
   implicit val hc = new HeaderCarrier()
```
jameshforster: Please remove the postData option from setupTarget and the relevant mock, this is no longer required.

jameshforster: Please remove the lazy val fakeRequest and instead extend the fakeRequestHelper trait.
Do the same for the buildRequest and executeTargetWithMockData methods (can be replaced using a method in the trait).

test/controllers/CalculationControllerTests/DisposalCostsActionSpec.scala
```diff
@@ -135,16 +97,8 @@ class DisposalCostsSpec extends UnitSpec with WithFakeApplication with MockitoSu
...
-        }
-
-        "have the value 1000 auto-filled into the input box" in {
-          document.getElementById("disposalCosts").attr("value") shouldEqual "1000"
-        }
+      s"have the title ${messages.question}" in {
+        document.getElementsByTag("title").text shouldBe messages.question
       }
     }
   }
```
jameshforster: Please add a test for the scenario where a request is made to the get without a valid session (can be found in the FakeRequestHelper) which should return a 303 redirecting to the session timeout page.

PR #697, opened 2016-11-01
Other Properties Spec update.

DavidWoodward37:


test/controllers/CalculationControllerTests/OtherPropertiesActionSpec.scala
```diff
@@ -0,0 +1,308 @@
...
+          status(result) shouldBe 303
+        }
+
+        "should redirect to the acquisitionDate page" in {
+          redirectLocation(result) shouldBe Some(s"${routes.AcquisitionDateController.acquisitionDate()}")
+        }
+      }
+    }
+
+    "for a personal rep" should {
```
jameshforster: Is there any value in testing based on customer type given that it has no impact on the methods called for a submit.

test/forms/nonResident/OtherPropertiesFormSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+
+      s"include an error message of '${messages.errorNegative}" in {
+        form.errors.head.message shouldEqual messages.errorNegative
+      }
+
+      s"include an error message of '${messages.errorDecimalPlaces}" in {
+        form.errors(1).message shouldEqual messages.errorDecimalPlaces
+      }
+    }
+  }
```
jameshforster: Could we have a test for proving that the amount validation is NOT run when a no is submitted?

test/views/nonResident/OtherPropertiesViewSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+
+      "have inputs using the id 'otherProperties'" in {
+        document.body().select("input[type=radio]").attr("id") should include("otherProperties")
+      }
+
+      "have inputs using the id otherPropertiesAmt" in {
+        document.body().select("input[type=number]").attr("id") shouldEqual ""
+      }
+    }
+  }
```
jameshforster: Could you add a test for the error summary (e.g. that it loads at all)

PR #698, opened 2016-11-01
Split unit tests for Other Reliefs Flat into individual Specs.

jameshforster:


test/forms/nonResident/OtherReliefsFormSpec.scala
```diff
@@ -0,0 +1,231 @@
...
+      val map = Map(
+        "isClaimingOtherReliefs" -> "Yes",
+        "otherReliefs" -> "-500.345"
+      )
+      lazy val form = otherReliefsForm(false).bind(map)
+
+      "return an invalid form with two errors" in {
+        form.errors.size shouldBe 2
+      }
+    }
```
TomDudley92: Could use a test for checking an error is produced when isClaimingOtherReliefs is not Yes or No

test/controllers/CalculationControllerTests/OtherReliefsFlatActionSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+      "return a 200 with a valid calculation result" in {
+        status(result) shouldBe 200
+      }
+
+      "load the otherReliefs flat page" in {
+        document.title() shouldBe messages.inputQuestion
+      }
+    }
+
+    "supplied with a pre-existing stored model and a loss" should {
```
TomDudley92: Can you remove "and a loss" please.

TomDudley92: Can you remove "and a loss" please.

test/forms/nonResident/OtherReliefsFormSpec.scala
TomDudley92: Could use a test for checking an error is produced when isClaimingOtherReliefs is not Yes or No

PR #719, opened 2016-11-10
(CGT-2143) Updated the content and routing for disposal value

euptonqa:


euptonqa:
@DavidWoodward37 updated thanks :)

PR #721, opened 2016-11-10
Updated the routing and content on the Disposal Value page.

DavidWoodward37:


test/views/nonResident/DisposalValueViewSpec.scala
```diff
@@ -44,7 +44,7 @@ class DisposalValueViewSpec extends UnitSpec with WithFakeApplication with Mocki
         }
 
         s"has a route to 'disposal-date'" in {
```
morgan-fernandes: Need to change to Sold for Less

PR #725, opened 2016-11-10
Added YesNo form helper.

jameshforster:


app/views/helpers/formYesNoRadio.scala.html
```diff
@@ -0,0 +1,63 @@
+@import uk.gov.hmrc.play.views.html.helpers.form
+@import play.api.i18n._
+
+@(field: Field, legend: String, legendClass: Option[String] = None, helpText: Option[String] = None)
+
+@{
+    if (helpText.isDefined && legendClass.isDefined) {
```
DavidWoodward37: Can you convert these to case matches please?

PR #741, opened 2016-11-16
Created query string request to calculate NRCGT total gain.

jameshforster:


app/connectors/CalculatorConnector.scala
```diff
@@ -57,6 +57,12 @@ trait CalculatorConnector {
     sessionCache.fetchAndGetEntry(key)
   }
 
+  def calculateTotalGain(totalGainAnswersModel: TotalGainAnswersModel)(implicit hc: HeaderCarrier): Future[Option[TotalGainResultsModel]] = {
+    http.GET[Option[TotalGainResultsModel]](s"$serviceUrl/capital-gains-calculator/non-resident/calculate-total-gain${
```
TomDudley92: Please add a ? after gain please. so it is gain?$

PR #746, opened 2016-11-17
Updated the Acquisition Date routes to match prototype

TomDudley92:


test/controllers/CalculationControllerTests/AcquisitionDateSpec.scala
```diff
@@ -45,6 +46,9 @@ class AcquisitionDateSpec extends UnitSpec with WithFakeApplication with Mockito
     when(mockCalcConnector.fetchAndGetFormData[AcquisitionDateModel](Matchers.anyString())(Matchers.any(), Matchers.any()))
       .thenReturn(Future.successful(getData))
 
+    when(mockCalcConnector.saveFormData[AcquisitionDateModel](Matchers.any(), Matchers.any())(Matchers.any(), Matchers.any()))
```
jameshforster: This shouldn't be necessary for the mock, given the returned null value is never used.

app/controllers/nonresident/AcquisitionDateController.scala
```diff
@@ -56,9 +57,24 @@ trait AcquisitionDateController extends FrontendController with ValidActiveSessi
     }
 
     def successAction(model: AcquisitionDateModel) = {
-      calcConnector.saveFormData(KeystoreKeys.acquisitionDate, model)
-      Future.successful(Redirect(routes.HowBecameOwnerController.howBecameOwner()))
+      for {
+        save <- calcConnector.saveFormData(KeystoreKeys.acquisitionDate, model)
```
jameshforster: Should be moved out of a for yield and the save method moved into the route request (blocking until complete, no asynchronous).

PR #747, opened 2016-11-17
Updated summary report to use new constructors and models.

jameshforster:


app/views/calculation/nonresident/summaryReport.scala.html
```diff
@@ -20,7 +20,7 @@
     <div class="header-container">
         <h1 class="heading-xlarge">
             <span class="pre-heading">@Messages("calc.summary.secondaryHeading")</span>
-            <b>&pound;@MoneyPounds(result.taxOwed).quantity</b>
+            <b>&pound;0.00</b>
```
euptonqa: Just wondering what's happening here?

PR #749, opened 2016-11-17
Create CheckYourAnswers view and controller.

jameshforster:


app/views/helpers/nonresident/questionAnswerTableRow.scala.html
```diff
@@ -0,0 +1,33 @@
...
+
+
+<table class="check-your-answers">
+    <thead>
+        <th class="heading-medium" colspan="2">@Messages("calc.checkYourAnswers.tableHeading")</th>
+        <th></th>
+    </thead>
+    <tbody>
+    @for(question <- questions) {
+        <tr id="@{s"${question.id}"}">
```
TomDudley92: You can remove these id's and use a data-question instead. @macgrewal

app/views/helpers/nonresident/questionAnswerTableRow.scala.html
```diff
@@ -0,0 +1,33 @@
...
+            <td id="@{s"${question.id}-data"}">
+            @{question.data match {
+                case string: String => string
+                case int: Int => int.toString
+                case bigDecimal: BigDecimal => s"£${MoneyPounds(bigDecimal, 2).quantity}"
+                case date: LocalDate => datePageFormatNoZero.format(date)
+                case boolean: Boolean => booleanToString(boolean)
+                case _ => ""
+            }}</td>
+            <td id="@{s"${question.id}-changeLink"}" class = "change-your-answer">
```
TomDudley92: With a screen reader this will only read as change and is not linked to question. 
It would be better to do this: `<td class = "change-your-answer"><a href="@question.link">Change<span class="visuallyhidden"> your response to the question @question.question</span></a></td>`

app/views/helpers/nonresident/questionAnswerTableRow.scala.html
```diff
@@ -0,0 +1,33 @@
...
+@(questions: Seq[(QuestionAnswerModel[Any])])
+
+
+<table class="check-your-answers">
+    <thead>
+        <th class="heading-medium" colspan="2">@Messages("calc.checkYourAnswers.tableHeading")</th>
+        <th></th>
+    </thead>
+    <tbody>
+    @for(question <- questions) {
```
TomDudley92: You can change this @for to a map instead.

PR #752, opened 2016-11-18
Updated the model, form, view and controllers and specs for rebased value and mandatory rebased value.

DavidWoodward37:


app/controllers/nonresident/RebasedValueController.scala
```diff
@@ -53,46 +53,47 @@ trait RebasedValueController extends FrontendController with ValidActiveSession
...
-        case (false, Some(data)) => Future.successful(Ok(calculation.nonresident.rebasedValue(rebasedValueForm.fill(data), acquisitionDate.hasAcquisitionDate)))
-        case (true, None) => Future.successful(Ok(calculation.nonresident.mandatoryRebasedValue(rebasedValueForm)))
-        case (false, None) => Future.successful(Ok(calculation.nonresident.rebasedValue(rebasedValueForm, acquisitionDate.hasAcquisitionDate)))
+    def routeRequest(routeToMandatory: Boolean, rebasedValue: Option[RebasedValueModel]): Future[Result] =
+      (routeToMandatory, rebasedValue) match {
+        case (x, Some(data)) if x => Future.successful(Ok(calculation.nonresident.mandatoryRebasedValue(rebasedValueForm(x).fill(data))))
+        case (y, Some(data)) if !y => Future.successful(Ok(calculation.nonresident.rebasedValue(rebasedValueForm(y).fill(data))))
+        case (q, None) if q => Future.successful(Ok(calculation.nonresident.mandatoryRebasedValue(rebasedValueForm(q))))
+        case (p, None) if !p => Future.successful(Ok(calculation.nonresident.rebasedValue(rebasedValueForm(p))))
```
jameshforster: Please revert to direct true/false matching rather than guard statements and use the passed in boolean variable instead of the matched value.

PR #760, opened 2016-11-21
Updated the disposal costs back link.

DavidWoodward37:


morgan-fernandes:
Needs to be worked on/merged once CGT-2188(Market Value when Sold/Gave Away) is merged

PR #770, opened 2016-11-23
Updated disposalValueRow constructor

miranda-hawkes:


app/constructors/nonresident/AnswersConstructor.scala
```diff
@@ -46,6 +45,16 @@ trait AnswersConstructor {
     val improvements = calculatorConnector.fetchAndGetFormData[ImprovementsModel](KeystoreKeys.improvements).map(data => data.get)
     val otherReliefsFlat = calculatorConnector.fetchAndGetFormData[OtherReliefsModel](KeystoreKeys.otherReliefsFlat)
 
+    def disposalValue(soldOrGivenAwayModel: SoldOrGivenAwayModel,
+                      soldForLessModel: Option[SoldForLessModel]): Future[DisposalValueModel] = (soldOrGivenAwayModel, soldForLessModel) match {
+      case (SoldOrGivenAwayModel(true), Some(SoldForLessModel(true))) =>
+        calculatorConnector.fetchAndGetFormData[DisposalValueModel](KeystoreKeys.disposalMarketValue).map(data => data.get)
+      case (SoldOrGivenAwayModel(false), Some(SoldForLessModel(false))) =>
```
jameshforster: SoldForLessModel needs to be set to _ if not sold (as it doesn't matter if there is any data there)

app/constructors/nonresident/AnswersConstructor.scala
```diff
@@ -49,10 +49,10 @@ trait AnswersConstructor {
                       soldForLessModel: Option[SoldForLessModel]): Future[DisposalValueModel] = (soldOrGivenAwayModel, soldForLessModel) match {
       case (SoldOrGivenAwayModel(true), Some(SoldForLessModel(true))) =>
         calculatorConnector.fetchAndGetFormData[DisposalValueModel](KeystoreKeys.disposalMarketValue).map(data => data.get)
-      case (SoldOrGivenAwayModel(false), Some(SoldForLessModel(false))) =>
-        calculatorConnector.fetchAndGetFormData[DisposalValueModel](KeystoreKeys.disposalMarketValue).map(data => data.get)
-      case _ =>
+      case (SoldOrGivenAwayModel(true), _) =>
```
jameshforster: This logic is incorrect.

Should either be SoldOrGivenAway(true), Some(_) => disposal value
or SoldOrGivenAway(false), _ => disposal market value

with the other case covered by the final case

Please update.

PR #780, opened 2016-11-28
Cgt 2255

DavidWoodward37:


test/controllers/CalculationControllerTests/PrivateResidenceReliefActionSpec.scala
```diff
@@ -357,7 +357,7 @@ class PrivateResidenceReliefActionSpec extends UnitSpec with WithFakeApplication
       }
 
       "redirect to the Allowable Losses page" in {
```
jameshforster: Change text to match test.

app/views/calculation/nonresident/privateResidenceRelief.scala.html
```diff
@@ -53,7 +53,7 @@
     }
 
 
-    <h1 class="heading-large">@Messages("calc.base.pageHeading")</h1>
+    <h1 class="heading-large">@Messages("calc.privateResidenceRelief.question")</h1>
```
jameshforster: Update class to xlarge

PR #785, opened 2016-11-29
Added PRR into the report.

DavidWoodward37:


app/controllers/nonresident/ReportController.scala
```diff
@@ -53,15 +53,20 @@ trait ReportController extends FrontendController with ValidActiveSession {
     calcConnector.calculateTotalGain(totalGainAnswersModel)
   }
 
+  def getPRRModel()(implicit hc: HeaderCarrier): Future[Option[PrivateResidenceReliefModel]] = {
```
jameshforster: Please add check here to see if PRR can be used in the first place. Otherwise return a None by default.

jameshforster:
Tests are failing. Please fix.

app/controllers/nonresident/ImprovementsController.scala
```diff
@@ -122,7 +121,7 @@ trait ImprovementsController extends FrontendController with ValidActiveSession
         val optionSeq = Seq(model.get.rebasedGain, model.get.timeApportionedGain).flatten
         val finalSeq = Seq(model.get.flatGain) ++ optionSeq
 
-        (finalSeq.forall(_ >= 0), skipPRR) match {
+        (!finalSeq.forall(_ <= 0), skipPRR) match {
```
jameshforster: Should be:

case (true, false) => Redirect(routes.PrivateResidenceReliefController.privateResidenceRelief())
          case (true, true) => Redirect(controllers.nonresident.routes.CustomerTypeController.customerType())
          case (_, _) => Redirect(routes.CheckYourAnswersController.checkYourAnswers())

PR #799, opened 2016-12-02
Removed ID for fieldset element in RadioInputGroup helper

euptonqa:
In addition to fixing affected ViewSpecs


euptonqa:
@TomDudley92

